PR #20, opened 2024-08-16
fallback to F32 for 1-dimensional or very small tensors

ddh0:
Currently the fallback data type is set to F16, even when a tensor is 1-dimensional, and even when the original model was in BF16.

Using anything less than f32 for 1-dimensional tensors can cause a big loss of quality, which is why GGML / llama.cpp always use f32 for their 1-dimensional tensors. (The only exception would be if every tensor in the model was originally F16, which I have never seen.)

Additionally this solves the issue of converting BF16 --> F16, which is not lossless.

For example, in the case of FLUX.1-dev, which is natively BF16:

- `original BF16 safetensors` --> `FP16 GGUF` --> `Q8_0 GGUF`

is worse than

- `original BF16 safetensors` --> `F32 gguf or F32 safetensors` --> `Q8_0 GGUF`

I hope my explanation makes sense and I hope this contribution is useful. Thanks!

### DIFF
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -83,7 +83,11 @@ def handle_tensors(args, writer, state_dict):
         n_dims = len(data.shape)
         data_qtype = args.qtype
         data_shape = data.shape
-        fallback = gguf.GGMLQuantizationType.F16
+        fallback = gguf.GGMLQuantizationType.F32
+
+        # NOTE: fallback data type is F32 because many models such as Flux
+        #       are published in BF16, and BF16 --> F16 is not lossless,
+        #       while BF16 --> F32 is lossless.
 
         if n_dims == 1: 
             # these barely take up space, just use F32 / F16.
@@ -123,12 +127,12 @@ def handle_tensors(args, writer, state_dict):
         try:
             data = gguf.quants.quantize(data, data_qtype)
         except gguf.QuantError as e:
-            tqdm.write(f"falling back to F16: {e}")
-            data_qtype = gguf.GGMLQuantizationType.F16
+            tqdm.write(f"falling back to F32: {e}")
+            data_qtype = gguf.GGMLQuantizationType.F32
             data = gguf.quants.quantize(data, data_qtype)
         except AttributeError as e:
-            tqdm.write(f"falling back to F16: {e}")
-            data_qtype = gguf.GGMLQuantizationType.F16
+            tqdm.write(f"falling back to F32: {e}")
+            data_qtype = gguf.GGMLQuantizationType.F32
             data = gguf.quants.quantize(data, data_qtype)
 
         new_name = key # do we need to rename?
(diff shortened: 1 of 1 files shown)


city96:
I did test this originally, but saw no actual discernible difference between FP16 and FP32 in the case of flux. I thought the FP32 in llama.cpp was mostly just because of what they use for internal data types?

Although you're probably right and this should be the default for future models. Won't requantize just yet since again, I didn't see any real quality loss and it's possible we make other format changes (i.e. metadata).

city96:
Also, changing the fallback to FP32 alltogether would result in all conv2d layers being also quantized to FP32, resulting in diffuser/future hybrid models with massive file sizes (I'm testing with 1.5 and it doubles the file size lol)

ddh0:
@city96 I just changed it so that it only falls back to F32 for 1D or very small tensors. Please take a look 👍

city96:
Perfect, thank you <3

One tiny nitpick: The FP32 logic should take precedence over the blacklist logic. i.e. keys matched by this that were decided to be FP32 will get put as FP16 by this being under the FP32 logic instead of above it:

```
        elif any([x in key for x in blacklist]) and ".weight" in key:
            data_qtype = fallback
```

ddh0:
Ah I see. Let me fix that

ddh0:
Done!

city96:
Perfect, thanks. Again, won't reupload just yet, probably will once we get more significant changes like metadata/etc.

On that note, here's some things that might be worth looking into if you feel like checking/testing stuff:

- Q8_0 with the old method is closer to whatever native ComfyUI outputs than Q8_0 with FP32 tensors for me. Maybe it's also being cast to FP16 along the way?
- Might be worth exploring what keys to add/remove from the blacklist to see if some can be quantized after all
- [this](https://github.com/city96/ComfyUI-GGUF/blob/51c95d7d7018e782b9a28729fa2d0499870da733/tools/convert.py#L69) line for the metadata is most likely wrong

(this is also more of a list for myself so I don't forget lol)

ddh0:
> Might be worth exploring what keys to add/remove from the blacklist to see if some can be quantized after all

You could forego the blacklist entirely and rely entirely on heuristics, like so:
```python
n_params = 1    # total number of elements: example: [3,4] -> 12
shape_sum = 1   # product of all dimension sizes: example: [3,4] -> 7
for dim_size in data_shape:
    n_params *= dim_size
    shape_sum += dim_size

if n_dims == 1:
    data_qtype = gguf.GGMLQuantizationType.F32
elif shape_sum >= 1024:
    data_qtype = gguf.GGMLQuantizationType.F32
elif n_params >= 65536:
    data_qtype = gguf.GGMLQuantizationType.F32
else:
    data_qtype = args.qtype
```

ddh0:
(Not 100% sure on those specific numbers as cutoffs, but they seem reasonable as a starting point)

PR #49, opened 2024-08-20
GGUF Text Encoder support

city96:
This should allow using the T5 encoder in other precisions than full FP16 and FP8. On a quick test with the Q4_0 unet the results between a BF16 T5 and a [Q5_K_M](https://huggingface.co/city96/t5-v1_1-xxl-encoder-gguf/blob/main/t5-v1_1-xxl-encoder-Q5_K_M.gguf) one were [almost the same](https://imgsli.com/Mjg5MjQy) - obviously this needs more testing but preliminary results look promising.

To use it simply replace the text encoder loader with the GGUF variant. You do **not** need GGUF versions of CLIP, it can load normal models with `.safetensors`/`.bin` extensions too.

T5_Q5_K_M

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -42,33 +42,6 @@ def gguf_sd_loader(path):
     print("\n")
     return sd
 
-class UnetLoaderGGUF:
-    @classmethod
-    def INPUT_TYPES(s):
-        unet_names = [x for x in folder_paths.get_filename_list("unet_gguf")]
-        return {
-            "required": {
-                "unet_name": (unet_names,),
-            }
-        }
-
-    RETURN_TYPES = ("MODEL",)
-    FUNCTION = "load_unet"
-    CATEGORY = "bootleg"
-    TITLE = "Unet Loader (GGUF)"
-
-    def load_unet(self, unet_name):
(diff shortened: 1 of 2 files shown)


city96:
@RandomGitUser321 Thanks for the tests, and do report back on your further findings.

>One test I really want to run is with q4 versions of both the t5 and flux

As you can see in the example image, using it with Q4_0 (or preferably Q4_K_M) is at least viable, so this should definitely help run it on weaker hardware.

Foul-Tarnished:
T5 FP8


T5 Q6_K


Quite a big diff

I didn't try FP16 (not sure I can even run it, and have slow internet)
so I don't know if it's more faithful or not

Foul-Tarnished:
Comparison of T5 (all on Flux base model as Q6_K)
flux-comparison
Euler, beta scheduler, 40 steps

Q6_K looks worse than FP8 for me (even though it has more details in background)
Q8 is quite good, but the hairstyle is not the same (bangs hair)

I would say FP16 > FP8 = Q8 > Q6_K
FP8 face and hairstyle is closer to FP16, but not the shirt, hmm

cc @city96

PR #63, opened 2024-08-23
Add support for SD1x and SDXL models

blepping:
this patch adds support for SD1.5 (tested, probably other SD1x and SD2x models) as well as SDXL. i know people say you shouldn't use quantization with conv-heavy models but the results with Q4_K_M (what i tested) are at least as good as BnB NF4 which is the only other option and doesn't support LoRAs.

not sure why the tensor shape thing was necessary but without it i was getting a failure due to the shape being made of `numpy.float64` items.

### DIFF
--- a/dequant.py
+++ b/dequant.py
@@ -7,15 +7,11 @@ from tqdm import tqdm
 def dequantize_tensor(tensor, dtype=torch.float16):
     data = torch.tensor(tensor.data)
     qtype = getattr(tensor, "tensor_type", None)
-    oshape = getattr(tensor, "tensor_shape", tensor.data.shape)
+    oshape = getattr(tensor, "tensor_shape", data.shape)
 
-    if qtype == None:
-        return data.to(dtype)
-    elif qtype == gguf.GGMLQuantizationType.F32:
-        return data.to(dtype)
-    elif qtype == gguf.GGMLQuantizationType.F16:
-        return data.to(dtype)
-    elif qtype in dequantize_functions:
+    if qtype in {None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}:
+        return data.to(dtype).view(oshape)
+    if qtype in dequantize_functions:
         # dequantize in fp16 then convert instead of keeping FP32
         out = dequantize(data, qtype, oshape, dtype=None)
         return out.to(dtype) if out.dtype != dtype else out # why is .to() not a no-op?
--- a/dequant.py
+++ b/dequant.py
@@ -4,11 +4,11 @@ import torch
 from tqdm import tqdm
 
 def dequantize_tensor(tensor, dtype=None, dequant_dtype=None):
-    data = tensor.data.clone().detach()
+    data = torch.tensor(tensor.data)
     qtype = getattr(tensor, "tensor_type", None)
     oshape = getattr(tensor, "tensor_shape", data.shape)
 
-    if qtype in [None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16]:
+    if qtype in (None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16):
         return data.to(dtype)
     elif qtype in dequantize_functions:
         dequant_dtype = dtype if dequant_dtype == "target" else dequant_dtype
--- a/ops.py
+++ b/ops.py
@@ -56,6 +56,7 @@ class GGMLLayer(torch.nn.Module):
     """
     This (should) be responsible for de-quantizing on the fly
     """
+    comfy_cast_weights = True
     dequant_dtype = None
     patch_dtype = None
 
@@ -63,6 +64,24 @@ class GGMLLayer(torch.nn.Module):
         super().__init__()
         self.weight = GGMLTensor(1, tensor_type=None, tensor_shape=None)
         self.bias = None
+
+    def _forward_operation(self, x, weight, bias):
+        raise NotImplementedError
+
+    def forward(self, x):
+        # lowvram hack
+        device = None
+        if self.weight.device != x.device:
+            device = self.weight.device
+            self.to(x.device)
+
+        weight, bias = self.get_weights(x.dtype)
+        x = self._forward_operation(x, weight, bias)
+        del weight, bias
+
+        if device:
+            self.to(device)
+        return x
 
     def _load_from_state_dict(self, state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs):
         for k,v in state_dict.items():
@@ -131,40 +150,17 @@ class GGMLLayer(torch.nn.Module):
         bias = self.get_weight(self.bias, dtype)
         return (weight, bias)
 
-class GGMLOpsLayer(GGMLLayer):
-    comfy_cast_weights = True
-
-    def __init__(self, *args, device=None, dtype=None, **kwargs):
-        super().__init__(device=device, dtype=dtype)
-
-    def _forward_operation(self, x, weight, bias):
-        raise NotImplementedError
-
-    def forward(self, x):
-        # lowvram hack
-        device = None
-        if self.weight.device != x.device:
-            device = self.weight.device
-            self.to(x.device)
-
-        weight, bias = self.get_weights(x.dtype)
-        x = self._forward_operation(x, weight, bias)
-        del weight, bias
-
-        if device:
-            self.to(device)
-        return x
 
 class GGMLOps(comfy.ops.manual_cast):
     """
     Dequantize weights on the fly before doing the compute
     """
-    class Linear(GGMLOpsLayer):
+    class Linear(GGMLLayer):
         _forward_operation = staticmethod(torch.nn.functional.linear)
 
-    class Conv2d(GGMLOpsLayer):
+    class Conv2d(GGMLLayer):
         def __init__(self, *args, device=None, dtype=None, **kwargs):
-            super().__init__(device=device, dtype=dtype)
+            super().__init__()
             _ = kwargs.pop("kernel_size", None)
             self._forward_operation = partial(staticmethod(torch.nn.functional.conv2d), **kwargs)
 
--- a/ops.py
+++ b/ops.py
@@ -16,7 +16,7 @@ class GGMLTensor(torch.Tensor):
         self.tensor_shape = tensor_shape
         self.patches = patches
 
-    def __new__(cls, *args, tensor_type, tensor_shape, patches=[], **kwargs):
+    def __new__(cls, *args, tensor_type, tensor_shape, tensor_data=None, patches=[], **kwargs):
         return super().__new__(cls, *args, **kwargs)
 
     def to(self, *args, **kwargs):
@@ -145,6 +145,29 @@ class GGMLOps(comfy.ops.manual_cast):
 
             weight, bias = self.get_weights(x.dtype)
             x = torch.nn.functional.linear(x, weight, bias)
+            del weight, bias
+
+            if device:
+                self.to(device)
+            return x
+
+    class Conv2d(GGMLLayer):
+        comfy_cast_weights = True
+
+        def __init__(self, *args, device=None, dtype=None, **kwargs):
+            super().__init__(device=device, dtype=dtype)
+            _ = kwargs.pop("kernel_size", None)
+            self.kwargs=kwargs
+
+        def forward(self, x):
+            # lowvram hack
+            device = None
+            if self.weight.device != x.device:
+                device = self.weight.device
+                self.to(x.device)
+
+            weight, bias = self.get_weights(x.dtype)
+            x = torch.nn.functional.conv2d(x, weight, bias, **self.kwargs)
             del weight, bias
 
             if device:
--- a/ops.py
+++ b/ops.py
@@ -1,5 +1,6 @@
 # (c) City96 || Apache-2.0 (apache.org/licenses/LICENSE-2.0)
 import torch
+from functools import partial
 
 import comfy.ops
 from .dequant import dequantize_tensor
@@ -14,7 +15,7 @@ class GGMLTensor(torch.Tensor):
         self.tensor_shape = tensor_shape
         self.patches = patches
 
-    def __new__(cls, *args, tensor_type, tensor_shape, tensor_data=None, patches=[], **kwargs):
+    def __new__(cls, *args, tensor_type, tensor_shape, patches=[], **kwargs):
         return super().__new__(cls, *args, **kwargs)
 
     def to(self, *args, **kwargs):
@@ -35,7 +36,7 @@ class GGMLTensor(torch.Tensor):
         try:
             return super().copy_(*args, **kwargs)
         except Exception as e:
-            print(f"ignoring 'copy_' on tensor")
+            print(f"ignoring 'copy_' on tensor: {e}")
 
     def __deepcopy__(self, *args, **kwargs):
         # Intel Arc fix, ref#50
@@ -130,53 +131,43 @@ class GGMLLayer(torch.nn.Module):
         bias = self.get_weight(self.bias, dtype)
         return (weight, bias)
 
+class GGMLOpsLayer(GGMLLayer):
+    comfy_cast_weights = True
+
+    def __init__(self, *args, device=None, dtype=None, **kwargs):
+        super().__init__(device=device, dtype=dtype)
+
+    def _forward_operation(self, x, weight, bias):
+        raise NotImplementedError
+
+    def forward(self, x):
+        # lowvram hack
+        device = None
+        if self.weight.device != x.device:
+            device = self.weight.device
+            self.to(x.device)
+
+        weight, bias = self.get_weights(x.dtype)
+        x = self._forward_operation(x, weight, bias)
+        del weight, bias
+
+        if device:
+            self.to(device)
+        return x
+
 class GGMLOps(comfy.ops.manual_cast):
     """
     Dequantize weights on the fly before doing the compute
     """
-    class Linear(GGMLLayer):
-        comfy_cast_weights = True
+    class Linear(GGMLOpsLayer):
+        _forward_operation = staticmethod(torch.nn.functional.linear)
 
-        def __init__(self, *args, device=None, dtype=None, **kwargs):
-            super().__init__(device=device, dtype=dtype)
-
-        def forward(self, x):
-            # lowvram hack
-            device = None
-            if self.weight.device != x.device:
-                device = self.weight.device
-                self.to(x.device)
-
-            weight, bias = self.get_weights(x.dtype)
-            x = torch.nn.functional.linear(x, weight, bias)
-            del weight, bias
-
-            if device:
-                self.to(device)
-            return x
-
-    class Conv2d(GGMLLayer):
-        comfy_cast_weights = True
-
+    class Conv2d(GGMLOpsLayer):
         def __init__(self, *args, device=None, dtype=None, **kwargs):
             super().__init__(device=device, dtype=dtype)
             _ = kwargs.pop("kernel_size", None)
-            self.kwargs=kwargs
+            self._forward_operation = partial(staticmethod(torch.nn.functional.conv2d), **kwargs)
 
-        def forward(self, x):
-            # lowvram hack
-            device = None
-            if self.weight.device != x.device:
-                device = self.weight.device
-                self.to(x.device)
-
-            weight, bias = self.get_weights(x.dtype)
-            x = torch.nn.functional.conv2d(x, weight, bias, **self.kwargs)
-            del weight, bias
-
-            if device:
-                self.to(device)
-            return x
 
 def move_patch_to_cuda(item, device):
     if isinstance(item, torch.Tensor):
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -69,6 +69,9 @@ def handle_tensors(args, writer, state_dict):
     # - do something about this being awful and hacky
 
     max_name_len = max([len(s) for s in state_dict.keys()]) + 4
+    if (max_name_len - 4) >= 64:
+        raise ValueError("Can only handle tensor names up to 63 characters!")
+    tensor_fixup = {}
     for key, data in tqdm(state_dict.items()):
         old_dtype = data.dtype
 
@@ -122,6 +125,11 @@ def handle_tensors(args, writer, state_dict):
                 #data = np.squeeze(data) # don't do this
                 data_qtype = gguf.GGMLQuantizationType.F16
 
+        if data_qtype != gguf.GGMLQuantizationType.F32 and n_dims >= 2 and n_params >= 256*256 and (n_params / 256).is_integer() and data.shape[0] >= 256 and data.shape[1] >= 256:
+            orig_shape=data.shape
+            orig_n_params=n_params
+            data = data.reshape(256+((n_params - 256*256) // 256),256)
+            tensor_fixup[key] = orig_shape
         try:
             data = gguf.quants.quantize(data, data_qtype)
         except gguf.QuantError as e:
@@ -139,6 +147,8 @@ def handle_tensors(args, writer, state_dict):
         tqdm.write(f"{f'%-{max_name_len}s' % f'{new_name}'} {old_dtype} --> {data_qtype.name}, shape = {shape_str}")
 
         writer.add_tensor(new_name, data, raw_dtype=data_qtype)
+    fixup_str = "\n".join(" ".join((k,)+tuple(str(dimval) for dimval in v)) for k,v in tensor_fixup.items())
+    writer.add_string("comfy.gguf.tensor_fixup", fixup_str)
 
 if __name__ == "__main__":
     args = parse_args()
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -9,7 +9,7 @@ from safetensors.torch import load_file
 
 QUANTIZATION_THRESHOLD = 1024
 REARRANGE_THRESHOLD = 512
-MAX_TENSOR_NAME_LENGTH = 63
+MAX_TENSOR_NAME_LENGTH = 127
 
 # Tuple of arch_name, match_lists.
 # Each item in match_lists is a tuple of keys that must match.
@@ -104,7 +104,6 @@ def handle_tensors(args, writer, state_dict):
     if max_name_len > MAX_TENSOR_NAME_LENGTH:
         bad_list = ", ".join(f"{key!r} ({namelen})" for key, namelen in name_lengths if namelen > MAX_TENSOR_NAME_LENGTH)
         raise ValueError(f"Can only handle tensor names up to {MAX_TENSOR_NAME_LENGTH} characters. Tensors exceeding the limit: {bad_list}")
-    tensor_fixup = {}
     for key, data in tqdm(state_dict.items()):
         old_dtype = data.dtype
 
@@ -157,7 +156,10 @@ def handle_tensors(args, writer, state_dict):
         ):
             orig_shape = data.shape
             data = data.reshape(n_params // 256, 256)
-            tensor_fixup[key] = orig_shape
+            writer.add_string(
+                f"comfy.gguf.orig_shape.{key}",
+                ", ".join(str(dim) for dim in orig_shape),
+            )
 
         try:
             data = gguf.quants.quantize(data, data_qtype)
@@ -172,12 +174,6 @@ def handle_tensors(args, writer, state_dict):
         tqdm.write(f"{f'%-{max_name_len + 4}s' % f'{new_name}'} {old_dtype} --> {data_qtype.name}, shape = {shape_str}")
 
         writer.add_tensor(new_name, data, raw_dtype=data_qtype)
-
-    if not tensor_fixup:
-        return
-
-    fixup_str = "\n".join(" ".join((k,)+tuple(str(dimval) for dimval in v)) for k,v in tensor_fixup.items())
-    writer.add_string("comfy.gguf.tensor_fixup", fixup_str)
 
 if __name__ == "__main__":
     args = parse_args()
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -156,10 +156,7 @@ def handle_tensors(args, writer, state_dict):
         ):
             orig_shape = data.shape
             data = data.reshape(n_params // 256, 256)
-            writer.add_string(
-                f"comfy.gguf.orig_shape.{key}",
-                ", ".join(str(dim) for dim in orig_shape),
-            )
+            writer.add_array(f"comfy.gguf.orig_shape.{key}", tuple(int(dim) for dim in orig_shape))
 
         try:
             data = gguf.quants.quantize(data, data_qtype)
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -13,6 +13,7 @@ def parse_args():
     parser = argparse.ArgumentParser(description="Generate F16 GGUF files from single UNET")
     parser.add_argument("--src", required=True, help="Source model ckpt file.")
     parser.add_argument("--dst", help="Output  unet gguf file.")
+    parser.add_argument("--force-arch", type=str, choices=("flux", "sd1", "sdxl"), default=None, help="Advanced option, allows forcing the model architecture. May be necessary for non-Flux models.")
     args = parser.parse_args()
 
     if not os.path.isfile(args.src):
@@ -39,11 +40,12 @@ def load_state_dict(path):
 
     return sd
 
-def load_model(path):
+def load_model(path, force_arch=None):
     state_dict = load_state_dict(path)
-
     # from ComfyUI model detection
-    if "transformer_blocks.0.attn.norm_added_k.weight" in state_dict:
+    if force_arch is not None:
+        arch = force_arch
+    elif "transformer_blocks.0.attn.norm_added_k.weight" in state_dict:
         arch = "flux"
         raise ValueError(f"The Diffusers UNET can not be used for this!")
     elif "double_blocks.0.img_attn.proj.weight" in state_dict:
@@ -54,7 +56,7 @@ def load_model(path):
         if "add_embedding.linear_1.weight" in state_dict:
             arch = "sdxl"
         else:
-            arch = "sd1" 
+            arch = "sd1"
     else:
         breakpoint()
         raise ValueError(f"Unknown model architecture!")
@@ -141,7 +143,7 @@ def handle_tensors(args, writer, state_dict):
 if __name__ == "__main__":
     args = parse_args()
     path = args.src
-    writer, state_dict = load_model(path)
+    writer, state_dict = load_model(path, force_arch=args.force_arch)
 
     writer.add_quantization_version(gguf.GGML_QUANT_VERSION)
     if next(iter(state_dict.values())).dtype == torch.bfloat16:
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -1,24 +1,55 @@
 # (c) City96 || Apache-2.0 (apache.org/licenses/LICENSE-2.0)
 import os
-import sys
 import torch
-import numpy as np
 import gguf # This needs to be the llama.cpp one specifically!
 import argparse
 from tqdm import tqdm
 
 from safetensors.torch import load_file
 
+QUANTIZATION_THRESHOLD = 1024
+REARRANGE_THRESHOLD = 512
+MAX_TENSOR_NAME_LENGTH = 63
+
+# Tuple of arch_name, match_lists.
+# Each item in match_lists is a tuple of keys that must match.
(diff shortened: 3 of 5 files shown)


city96:
The model arch should always be auto-detected so allowing people to force it like that is a bad idea IMO - you should instead add whatever keys your SDXL checkpoint comes with (i.e. the LDM ones) to the model detect if switch.

You'd also have to change the logic for _M quants to handle the different ffn names otherwise _M and _S quants would basically be the same or very close in size since they'd default to the base qtype with no logic.

I can make these changes and credit you in the commit message if you're not too sure about the logic on that later part, or you can make the requested changes yourself. Either works.

blepping:
unfortunately, this isn't working as well as it seemed at first. due to a bug in reporting the fallback quant type (fixed in the second commit) it wasn't actually showing that it was falling back to F16. most of the tensors in SD1 were not compatible with the quant block size requirements so the upshot is not much was actually getting quantized.

the total number of elements actually almost always is compatible and since we don't care about actually performing operations on the quantized data it is possible to rearrange the dimensions to meet the requirements. however, that requires storing metadata with the original sizes which is sort of hacky. i have a proof of concept working (-ish, actually trying to generate stuff doesn't work).

not sure if you'd even be interested in that sort of approach.

city96:
>most of the tensors in SD1 were not compatible with the quant block size requirements so the upshot is not much was actually getting quantized.

Yeah this is the main problem. I saw a reduction from like 1.7 to 1.5GBs on SD1.2 and decided it was not worth it. I might make a proof of concept on some useless model like that or SDXL v0.9 to show people what performance/file sizes look like but I doubt it'd be very useful as a quant.

But hey, happy at least someone is experimenting.

> it is possible to rearrange the dimensions to meet the requirements.

I think flattening the tensors and storing the original sizes for conv2d is a *bit* too hacky. I left a comment [here](https://github.com/city96/ComfyUI-GGUF/blob/377862a06ddd7b0f599a768a305497be8068c07e/tools/convert.py#L120) that most likely only succeeds in making me look deranged, but that's what that was about lmao.

city96:
Also, I'm trying to get the original llama.cpp code + cuda kernels working, so anything weird like padding / flattening the weights would add a *lot* of overhead on top of that I'd assume.

blepping:
> Yeah this is the main problem. I saw a reduction from like 1.7 to 1.5GBs on SD1.2 and decided it was not worth it.

unfortunately not really worth it when barely anything gets quantized. big reduction when the tensors are compatible though:

```plaintext
llama_model_quantize_internal: model size  =  1640.25 MB
llama_model_quantize_internal: quant size  =   467.97 MB
llama_model_quantize_internal: WARNING: 2 of 282 tensor(s) required fallback quantization
```

Q4_K SD1.5

> I think flattening the tensors and storing the original sizes for conv2d is a bit too hacky.

fair enough. maybe i'll have to make my own Comfy GGUF with blackjack and hookers!

> Also, I'm trying to get the original llama.cpp code + cuda kernels working, so anything weird like padding / flattening the weights would add a lot of overhead on top of that I'd assume.

shouldn't need padding. i don't think it would be a problem unless you want to do stuff like math operations on the quantized data (as opposed to just dequantizing it into a tensor). so something like GGML's row-based quant mat mul kernel would have an issue but i think the dequantization kernels would be fine.

city96:
It's definitely worth it if we can get it to work without messing up the standard too much lol.

Wonder if it's worth asking the original llama.cpp team for their opinion on supporting conv2d at least at the quantization/dequantization level without having to flatten it first into a 2D tensor. Tagging @compilade

(I *was* going to try to use the cuda mat mul stuff but looks like that's a no-go with the fused qkv heads. I'm also way out of my depth on that one lol.)

blepping:
so, i actually got it working so i figured i'd push the changes. obviously just proof of concept at the moment, aside from the unavoidable hacks the code is pretty messy and can be cleaned up.

the error with conv not implemented in Comfy ops was really not obvious (except in hindsight). don't ask me how long it took to realize why stuff seemed to be trying to use the raw quantized tensor data.

> Wonder if it's worth asking the original llama.cpp team for their opinion on supporting conv2d at least at the quantization/dequantization level without having to flatten it first into a 2D tensor.

it isn't really so much a conv2d thing as an odd tensor dimensions thing. i'm not sure there's really a way around it with how block sizes in quants work.

>  I'm also way out of my depth on that one lol.

probably makes sense to do some performance measuring to see if that much time is even being spent dequantizing stuff. it might not make a huge difference even if you could speed it up a lot.

city96:
>the error with conv not implemented in Comfy ops was really not obvious (except in hindsight). don't ask me how long it took to realize why stuff seemed to be trying to use the raw quantized tensor data.

Yeah, I wasted more hours than I care to admin on the ops thing. At least it didn't strip the tensor attributes randomly for you.

>probably makes sense to do some performance measuring to see if that much time is even being spent dequantizing stuff. it might not make a huge difference even if you could speed it up a lot.

The current kernels can do between 500 and 1500it/s for a given weight. Flux has 169 keys quantized. If I'm doing the math right that's still a pretty significant slowdown, and probably compute overhead from not having fused operations. I only have CPU code (mostly) working but dequant performance went from 6it/s to 40it/s on a random weight I was testing on.

>the GGML C library is hard coded to use a size 64 buffer for tensor names.

You can change/override GGML_MAX_NAME in `ggml/include/ggml.h` or via a build flag.

blepping:
> You can change/override GGML_MAX_NAME

yeah, i saw that. patching that as well is possible, but patching less stuff seems better in general. encoding the metadata in the tensor names seems worse from a hackiness perspective also, it's just easier. but eh, i already did the work now.

anyway, not that you've seen these changes do you think it fits in the category of

> It's definitely worth it if we can get it to work without messing up the standard too much

or not? the general approach, that is. the current state obviously needs cleanup and tuning before it would be suitable to merge.

compilade:
I think the GGUF format by itself *can* handle quant blocks which span multiple rows (because all it stores about tensors are their shape, type and a blob of data), but this is not currently done because having blocks which begin and end at row boundaries helps a lot with making the matrix multiplication kernels more efficient.

But if using *only* dequantization-based kernels, then this is not really a problem.

The only problem would be the various checks in the (de)quantization functions which assume blocks never cross row boundaries.

city96:
@compilade looks like the main issue when disabling the check is with the tensor shapes. It seems like the order is the exact opposite of what it should be, almost like 4D tensors are ignored by some part of the logic.

image


Swapping around the shapes with some really hacky code makes it pass even without the checks disabled, but the output is garbage since I did this halfway through the quantize function so I'm pretty sure the element count is invalid at this stage.

```
tensor->ne[3] = 1;
tensor->ne[2] = 1;
tensor->ne[1] = ne0 * ne1 * ne2;
tensor->ne[0] = ne3;
```

In the F16 one they're stored correctly so it's not an issue at the convert stage I don't think:

image

(Also, feel free to tell me to open an issue on the llama.cpp repo if you think that's more suited)

compilade:
@city96 the order of the dimensions of the convolution tensors is not wrong, I think; unless I'm misunderstanding something, this is a 3x3 convolution kernel (or many?). The shapes also look "reversed-but-aren't" in the `safetensors` previews I checked, with the size of the rows (the last dimension) only being 3 or 1.

(Of course, by now you know that shapes in `ggml` are reversed compared to shapes in Numpy/PyTorch (the row size (aka the number of columns) is `ne[0]` in `ggml` while it is `shape[-1]` in Numpy/PyTorch))

For quantization with blocks spanning multiple rows, I think the most appropriate reshape migjt be to make the tensor 1D (although what you did seems reasonable), which then becomes 2D like `(-1, block_size)` when quantizing block-wise. (Which allows keeping the "row size is a multiple of block size" checks)

As long as the tensors are reshaped correctly after dequantization, there should not be problems (I think?).

What I know is that the quantization functions have a really bad behavior when passed tensors which don't have a row size divisible by the block size. With Mamba-130M, there was a tensor with 48 elements per row, which, when quantized with `Q4_0` (which has blocks of 32 elements), the result was not good *at all*. Not sure why, but likely some elements were ignored or maybe there were out-of-bound reads. Anyway, it did not produce a good result.

I think reshaping and making sure the new row size is exactly divisible by the block size of the chosen quant type should not cause this problem, though.

blepping:
> For quantization with blocks spanning multiple rows, I think the most appropriate reshape might be to make the tensor 1D 

i tried that. seems like there's some other logic that would need to be disabled, because 1D tensors don't seem to get quantized.

compilade:
> because 1D tensors don't seem to get quantized

Oh, right, this is true.

Then the 2D block size reshape should be more appropriate.

But be careful with keeping the strides in `tensor->nb[1]` consistent with the new `tensor->ne[0]` (e.g. `tensor->nb[1] = tensor->nb[0] * tensor->ne[0]`, assuming `tensor->nb[0]` is already `sizeof(float)` in this to-be-quantized tensor (which it should be if the tensor was already contiguous)). That could probably really mess up what data is quantized otherwise.

city96:
What I'm mostly questioning is why the 3x3 kernel shape is inverted compared to the 1x1 kernel right above it in the debug print.

compilade:
> What I'm mostly questioning is why the 3x3 kernel shape is inverted compared to the 1x1 kernel right above it in the debug print.

Likely because the convert script uses `squeeze()`, which removes the `1`-sized dimensions. I think `squeeze` should not be used in this case.

city96:
Ah, that would make sense. So it inadvertently works since it gets flattened to a 1D/2D one. I should probably redo the FP16 base as well with an unmodified convert script.

blepping:
> Then the 2D block size reshape should be more appropriate

i think i'm doing what you suggested (math isn't my forte though). my approach was to reshape to `(256 + ((n_elements - 256 * 256) // 256), 256)`. from the GGML side the result would look like `[  256, 14400,     1,     1]`

city96:
Yeah, okay, I was definitely flattening those somewhere in that F16. A bit hard to tell if it's the wrong way around if those dims are 1,1 anyway and the other two are 320x320 lol. Looks like that's all good on a fresh convert.

image

I think the reshape solution should be fine. Wonder if we can write the real shape into the output gguf instead of adding our own key (maybe even reshape directly in c++?), or if that would again be too hacky and cause issues down the line.

blepping:
> I think the reshape solution should be fine. Wonder if we can write the real shape into the output gguf instead of adding our own key (maybe even reshape directly in c++?), or if that would again be too hacky and cause issues down the line.

i'd think you would want to keep the llama.cpp patches as small/simple as possible. the larger and more complex they are, the more effort it's going to take to maintain them.

the writing the original shape part can't happen on the `convert.py` side because... we can already easily do that, it'll just prevent stuff from getting quantized. :)

is storing some metadata in a key really that hacky? it's using the built in features of the format, can't easily break, pretty much future proof and doesn't increase the maintenance burden. obviously it would be better if that stuff wasn't necessary, but it feels to me like having to reshape stuff is the hacky (and apparently mostly unavoidable) part, not the metadata.

ops.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, *args, tensor_type, tensor_shape, patches=[], **kwargs):
         self.tensor_shape = tensor_shape
         self.patches = patches
 
-    def __new__(cls, *args, tensor_type, tensor_shape, patches=[], **kwargs):
+    def __new__(cls, *args, tensor_type, tensor_shape, tensor_data=None, patches=[], **kwargs):
```
city96, line 19: This is probably left over from testing

ops.py
```diff
@@ -150,3 +150,26 @@ def forward(self, x):
...
+    class Conv2d(GGMLLayer):
+        comfy_cast_weights = True
+
+        def __init__(self, *args, device=None, dtype=None, **kwargs):
+            super().__init__(device=device, dtype=dtype)
+            _ = kwargs.pop("kernel_size", None)
+            self.kwargs=kwargs
+
+        def forward(self, x):
+            # lowvram hack
```
city96, line 163: I can deduplicate this part and clean it up once we merge it, I had it originally when I was adding T5 but ended up not needing it.

tools/convert.py
```diff
@@ -13,6 +13,7 @@ def parse_args():
     parser = argparse.ArgumentParser(description="Generate F16 GGUF files from single UNET")
     parser.add_argument("--src", required=True, help="Source model ckpt file.")
     parser.add_argument("--dst", help="Output  unet gguf file.")
+    parser.add_argument("--force-arch", type=str, choices=("flux", "sd1", "sdxl"), default=None, help="Advanced option, allows forcing the model architecture. May be necessary for non-Flux models.")
```
city96, line 16: This should definitely be removed and the right keys should be added to the check instead.

tools/convert.py
```diff
@@ -39,11 +40,12 @@ def load_state_dict(path):
...
 
-def load_model(path):
+def load_model(path, force_arch=None):
     state_dict = load_state_dict(path)
-
     # from ComfyUI model detection
-    if "transformer_blocks.0.attn.norm_added_k.weight" in state_dict:
+    if force_arch is not None:
+        arch = force_arch
+    elif "transformer_blocks.0.attn.norm_added_k.weight" in state_dict:
```
city96, line 48: Add the proper key here for diffusers instead of the override. Something like `down_blocks.1.attentions.1.transformer_blocks.0.attn1.to_k.weight` should work.

tools/convert.py
```diff
@@ -54,7 +56,7 @@ def load_model(path):
         if "add_embedding.linear_1.weight" in state_dict:
```
city96, line 56: These should also be changed/fixed. Add `label_emb.0.0.weight` as a key to check to detect SDXL.

tools/convert.py
```diff
@@ -67,6 +69,9 @@ def handle_tensors(args, writer, state_dict):
     # - do something about this being awful and hacky
 
     max_name_len = max([len(s) for s in state_dict.keys()]) + 4
+    if (max_name_len - 4) >= 64:
```
city96, line 72: I think the max name len was just there to make the print look nice lol. I think you could readd this further down like this since it's cleaner and actually prints the key that failed the check:

https://github.com/city96/ComfyUI-GGUF/blob/165212919e1c682f6ff740f9b69d968634b6d89f/tools/convert.py#L147

tools/lcpp.patch
```diff
@@ -43,7 +46,7 @@ index 5ab65ea9..44fe3fe6 100644
      // get hparams kv
      ml.get_key(LLM_KV_VOCAB_SIZE, hparams.n_vocab, false) || ml.get_arr_n(LLM_KV_TOKENIZER_LIST, hparams.n_vocab);
  
-@@ -15854,26 +15867,40 @@ static ggml_type llama_tensor_get_type(quantize_state_internal & qs, ggml_type n
+@@ -15854,26 +15870,40 @@ static ggml_type llama_tensor_get_type(quantize_state_internal & qs, ggml_type n
```
city96, line 49: We'll have to adjust the logic below this line for SD1/SDXL/etc so the in/out/embed/etc keys get kept in the right precision. I can double check this once we merge so this is more of a note to myself though you can add it like this as a start:

```cpp
    if ( // KEEP IN FP32
            (name == tn(LLM_TENSOR_OUTPUT, "weight")) ||
            (!qs.has_output && name == tn(LLM_TENSOR_TOKEN_EMBD, "weight")) ||
            // flux
            (name.find("img_in.") != std::string::npos) ||
            (name.find("time_in.in_layer.") != std::string::npos) ||
            (name.find("vector_in.in_layer.") != std::string::npos) ||
            (name.find("guidance_in.in_layer.") != std::string::npos) ||
            (name.find("final_layer.linear.") != std::string::npos) ||
            // SD
            (name.find("conv_in.") != std::string::npos) || 
            (name.find("conv_out.") != std::string::npos) ||
            (name == "input_blocks.0.0.weight") ||
            (name == "out.2.weight")
        ) {
```

```cpp
    } else if ( // KEEP IN FP16
            (name == "token_embd.weight") ||
            // flux
            (name.find("txt_in.") != std::string::npos) ||
            (name.find("time_in.") != std::string::npos) ||
            (name.find("vector_in.") != std::string::npos) ||
            (name.find("guidance_in.") != std::string::npos) ||
            (name.find("final_layer.") != std::string::npos) ||
            // SD
            (name.find("time_embed.") != std::string::npos) ||
            (name.find("time_embedding.") != std::string::npos) ||
            (name.find("label_emb.") != std::string::npos) ||
            (name.find("add_embedding.") != std::string::npos)
        ) {
```

tools/convert.py
```diff
@@ -120,6 +125,11 @@ def handle_tensors(args, writer, state_dict):
                 #data = np.squeeze(data) # don't do this
                 data_qtype = gguf.GGMLQuantizationType.F16
 
+        if data_qtype != gguf.GGMLQuantizationType.F32 and n_dims >= 2 and n_params >= 256*256 and (n_params / 256).is_integer() and data.shape[0] >= 256 and data.shape[1] >= 256:
```
city96, line 128: I think the n_dims == 4 check for FP16 above this can safely be removed since we're not doing any quantization in this script (i.e. not used anymore).

city96 (changes requested):
Okay, added my 5c with some suggested changes. LMK what you think.
Also tried to suggest a possible strategy for what weights to keep in higher precision for our llama.cpp patch part.

city96:
Huh, interesting. If you leave the original shape, even the FP32 conversion will destroy the weight for some reason if the shape is 4D.

image

While leaving it as FP16 (no conversion) works fine:

image

This is with all llama.cpp checks enabled. I guess you can move those two layers down to the FP16 part or force them as FP32 in the conversion script though it still feels like a bug.

ops.py
blepping: you're correct. when i didn't understand why the conv stuff was using the tensor data, i tried storing it separately.

i will fix this stuff and clean it up. i was just holding off to wait and see if you were interested in the approach in general/if this had a chance of being merged before i put more time into it. i'll probably get a chance to work on it tomorrow.

you probably already know this, but you can edit the branch a pull targets before merging it. i've found it useful to basically take ownership of a pull that way: create a branch, merge the person's pull into it and then i can make my own changes (and possibly own pull request) before i merge it into the main branch and they'll still get appropriate credit. it may make sense to do that after i've completed most of my changes. since you seem to have a better grasp of how to detect the model types/adjust the quant heuristics, it might be easier for you to just do yourself rather than directing me.

just a possible approach, definitely don't have to do it that way or anything.

tools/convert.py
blepping: this heuristic should be adjusted also, 256*256 is probably to small to be worth quantizing. also it's only checking if we _can_ rearrange stuff but not if we must, so it's probably doing that more aggressively than is actually necessary.

ops.py
blepping: you mean how the conv thing duplicates the linear layer code? yeah, definitely, was just to make sure it actually worked. there should be a better way to do it. it seemed weird/possibly wrong how it was inheriting from the normal linear layer.

compilade:
> If you leave the original shape, even the FP32 conversion will destroy the weight for some reason if the shape is 4D.

This is definitely a bug. It seems like `llama_model_quantize_internal` basically ignores `tensor->ne[3]`:

https://github.com/ggerganov/llama.cpp/blob/e11bd856d538e44d24d8cad4b0381fba0984d162/src/llama.cpp#L17062-L17075

ops.py
city96: >i will fix this stuff and clean it up. i was just holding off to wait and see if you were interested in the approach in general/if this had a chance of being merged before i put more time into it. i'll probably get a chance to work on it tomorrow.

Oh yeah, I definitely think this is worth merging now, and in hindsight I do agree that your solution of using the metadata in the gguf file is the less hacky approach. If we ever find a cleaner way due to some changes upstream we can still support loading both as it's not a massive overhead or anything.

You did a great job, the 1GB VRAM crowd will love this lol.

image

>you probably already know this, but you can edit the branch a pull targets before merging it.

Yeah just figured it'd be easier if you made all the changes you wanted/had planned first but I'm open to doing either. You can leave the model logic out and ping me when you're done if you'd prefer that.

blepping: > Yeah just figured it'd be easier if you made all the changes you wanted/had planned first

yes, absolutely, that's what i meant. not right now, but possibly after i've completed my changes and cleaned stuff up.

> You did a great job, the 1GB VRAM crowd will love this lol.

thanks!

city96: Sounds good, no rush, we may also want to wait for upstream llama.cpp for that FP32 bug mentioned below if we don't want to work around it.

Once it's all done I'll pop some example models like SDXLv0.9 and SD1.2 on HF to show the general size reduction while encouraging people to make their own quants with actually useful models.

blepping:
i think this now addresses most of the requested changes.

* fixed architecture detection/got rid of the force architecture flag (also made it easier to add more architectures or heuristics in the future)
* fixed the logic for rearranging stuff which was wrong and also needlessly complicated. rearranging now hits some smaller tensors like `3, 3, 320, 320` - this was necessary to make quantizing with non-K quants work.
* refactored the ops to reduce code duplication.
* got rid of the dimensions 4 F16 stuff in convert like you suggested.
* some other cleanups and code style stuff like removing unused imports.

currently does not include a workaround for the GGML F32 thing - it would be easy to add though. let me know if you want it and i can add that.

the one thing i didn't do was add your suggested heuristics changes to the `llama.cpp` patch. looking at that code, i think it would probably be better to make a copy of that `llama_tensor_get_type` function and put the image model logic in it (and remove all the unnecessary LLM stuff). that way you can just add an architecture check and delegation to the image model stuff to the top of the original function. it's likely that function will get updated frequently with new heuristics, new models, etc and if you're changing stuff in it then you're going to have to keep fixing your patch. just a suggestion of course, not telling you what you should do or anything like that. if you do want to do it that way, let me know and i can refactor it that way.

i tested these changes with SD1.5 and SDXL _non diffusers_ models (i don't have any diffusers models around).

i'm bad at naming stuff so if you want to rename any of the functions/classes or the added GGUF key, let me know. i don't really have any preference so i can just change it to whatever you want.

what do you think of the current state?

tools/convert.py
blepping: i wasn't sure whether you were referring to Flux or SD1 or SDXL here - the quoted section in the code is Flux but it would make more sense if you meant a non-Flux key. anyway, i didn't do anything with this yet.

dequant.py
```diff
@@ -4,9 +4,9 @@
 from tqdm import tqdm
 
 def dequantize_tensor(tensor, dtype=None, dequant_dtype=None):
-    data = torch.tensor(tensor.data)
+    data = tensor.data.clone().detach()
```
city96, line 7: I was doing this on purpose so the extra attributes don't have to be copied for all the ops inside the dequant kernels. https://github.com/city96/ComfyUI-GGUF/issues/66
 .clone() is disabled and .detach() also doesn't do anything since we don't have gradients so you could just put "tensor.data" and it'd be the same lol.

ops.py
```diff
@@ -130,30 +131,43 @@ def get_weights(self, dtype=torch.float16):
         bias = self.get_weight(self.bias, dtype)
         return (weight, bias)
 
+class GGMLOpsLayer(GGMLLayer):
```
city96, line 134: This is a nice way of doing it. Think they can just be part of `GGMLLayer` though, don't think that class will ever be used for anything that's not just, well, that.

tools/convert.py
```diff
@@ -97,47 +135,50 @@ def handle_tensors(args, writer, state_dict):
...
+                # one-dimensional tensors should be kept in F32
+                # also speeds up inference due to not dequantizing
+                data_qtype = gguf.GGMLQuantizationType.F32
+
+            elif n_params <= QUANTIZATION_THRESHOLD:
+                # very small tensors
+                data_qtype = gguf.GGMLQuantizationType.F32
+
+            elif ".weight" in key and any(x in key for x in blacklist):
+                data_qtype = gguf.GGMLQuantizationType.F32
```
city96, line 151: I think llama.cpp converts them to FP32 anyway during the quantization, which was the reason this was here. No one will use the FP16 checkpoints anyway just figured it's nicer if it matches since it barely increases the checkpoint size. We can leave them as fp16 though, it's fine.

tools/convert.py
```diff
@@ -97,47 +135,50 @@ def handle_tensors(args, writer, state_dict):
...
+        tqdm.write(f"{f'%-{max_name_len + 4}s' % f'{new_name}'} {old_dtype} --> {data_qtype.name}, shape = {shape_str}")
 
         writer.add_tensor(new_name, data, raw_dtype=data_qtype)
 
+    if not tensor_fixup:
+        return
+
+    fixup_str = "\n".join(" ".join((k,)+tuple(str(dimval) for dimval in v)) for k,v in tensor_fixup.items())
+    writer.add_string("comfy.gguf.tensor_fixup", fixup_str)
+
```
city96, line 181: You know, I was wondering. What if we add a separate fixup str for each key that needed to be fixed instead of a single string. We *may* need to adjust the max key length in llama.cpp for that but I think it may be nicer/easier to see when you look at the metadata on some place like HF.

Then you can just do something like this for each key:

```py
writer.add_string(f"comfy.gguf.orig_shape.{key}",  str(orig_shape))
```
```py
orig_shape = reader.get_field(f"comfy.gguf.orig_shape.{key}") # in a try/catch
```

city96:
I forgot to add the actual comment body but most of it looks good now. Added some comments with some explanations for some of the stuff + an idea I had for the fixup keys. I like the idea of a separate function in the cpp code, then we can just add a switch somewhere based on the arch and have a list of image models. I'll add that after the merge, thanks for the idea.

tools/convert.py
blepping: >  The key names aren't meant to store attributes/information like that IMO.

not to be argumentative, but isn't `f"comfy.gguf.orig_shape.{key}"` also storing metadata in the key name? except rather than it being the shape information directly, the metadata is the original _key_ instead of the original _shape_.

i'll admit i didn't give any consideration to stuff like how it will get displayed on HF though so that may be enough of a reason by itself.

> This is a valid concern, though that key name is under an `ifndef` so they can probably rebuild with `-DGGML_MAX_NAME=128` ..?

that's true, however the error you get is not at all intuitive. it's something like the tensor data isn't within the file (i forget the exact wording). it's definitely not anything along the lines of "the tensor name exceeds what was defined in `GGML_MAX_NAME`". took me a while to figure out what was happening, i originally though it just didn't like the characters i was using in the name.

***

anyway, seems like the verdict is to change it to the `comfy.gguf.orig_shape.ORIG_KEY` format? juuust double checking before i go ahead and mess with it.

city96: >not to be argumentative, but isn't f"comfy.gguf.orig_shape.{key}" also storing metadata in the key name

Touche.

There is also a `writer.add_array` function which would allow us to not do the whole string to list thing. I guess that's also another reason to do it this way though? It just takes/returns a list of ints as far as I can tell.

>how it will get displayed on HF

I'd assume the same way as the chat template on models, which looks pretty stupid without the newlines resolved sadly:

image

>anyway, seems like the verdict is to change it to the comfy.gguf.orig_shape.ORIG_KEY format? juuust double checking before i go ahead and mess with it.

Yep I think that's for the best.

blepping: how about this?

city96: Looks good. Any reason to use strings instead of `writer.add_array`?

blepping: > Any reason to use strings instead of writer.add_array?

i haven't messed with arrays. it would complicate extracting the values/checking the types, you also can't control what the internal type is for array items: GGUF will take a guess. looks like for `int` it will use a signed 32bit type for items. not sure how likely it is, but that means if a tensor's original dimension is over 2,147,483,648 stuff will break.

do you need me to change it?

city96: Not sure if this would break anything else but to me it looks like converting it back is a lot cleaner/easier when it's an array:

```py
test = reader.get_field("tokenizer.ggml.token_type")
# check if exists
shape = torch.Size(test.data)
```

I think if we hit the 32 bit int limit in one dimension on a tensor we'll have more pressing things to worry about.

blepping: i don't think your example would work, `.data` in a `ReaderField` doesn't lay the data out like that. also `data` isn't the the data itself, it's a list of indexes into `parts` for the parts of data that might make up a field. for example, an array will have an item for the key length, then the key data, then the array length, then the array data.

i will change it, but the code is going to be more complicated and possibly more fragile also.

city96: Huh, that example worked with the random LLM model I read the attributes out of and returned a valid object.

city96: Okay the tokenizer stuff must be added in some other way, let me look at it lol.

city96: Nvm I'm stupid it really was just the memory addresses lol.

city96: There must be a proper way to use it though if the web one can read it just fine.

image

blepping: yeah, it's weird because a field with `320, 4, 3, 3` actually looks like this:

```plaintext
ReaderField(offset=144, 
  name='comfy.gguf.orig_shape.input_blocks.0.0.weight', 
  parts=[
    memmap([45], dtype=uint64), <-- length of key
    memmap([ 99, 111, 109, 102, 121,  46, 103, 103, 117, 102,  46, 111, 114, 105, 103,  95, 115, 104,  97, 112, 101,  46, 105, 110, 112, 117, 116,  95,  98, 108, 111,  99, 107, 115,  46,  48,  46,  48,  46, 119, 101, 105, 103, 104, 116], dtype=uint8), <-- key data
    memmap([9], dtype=uint32), <-- type (ARRAY)
    memmap([5], dtype=uint32), <-- subtype (INT32)
    memmap([4], dtype=uint64), <-- array length
    memmap([320], dtype=int32), <-- array item 0 (index 5 in parts)
    memmap([4], dtype=int32), 
    memmap([3], dtype=int32), 
    memmap([3], dtype=int32) <-- array item 3 (index 8 in parts)
  ], 
  data=[5, 6, 7, 8], <-- indexes into parts for the actual data
  types=[<GGUFValueType.ARRAY: 9>, <GGUFValueType.INT32: 5>] <-- types that comprise the field
)
```

anyway, i changed it. took a while to figure out since arrays are quite weird to deal with.

city96: You're the man. Now that's what I call clean. I'll give it a test, finish up the C++ parts and merge it if I don't find any issues.

city96 (approved):

tools/convert.py
blepping: sounds good! let me know if you need any other changes/have questions/etc.

city96: Okay, I've pushed a completely rewritten convert patch which touches as little of the llama.cpp stuff as possible. As far as I can tell it's 1:1 to the old one with flux (sidenote, kind of amazing that the llama.cpp team managed to get this to be *literally* bit for bit reproducible lol.)

image

If you want you can pull in the changes here (maybe test them) and we can merge. If not I'll just resolve the merge conflict on the patch by keeping my one.

blepping: tested with SD1.5, seems to work. small difference in sizes:

```plaintext
=== BEFORE
llama_model_quantize_internal: model size  =  1639.41 MB
llama_model_quantize_internal: quant size  =   467.09 MB

=== AFTER
llama_model_quantize_internal: model size  =  1639.41 MB
llama_model_quantize_internal: quant size  =   503.95 MB
```

there are some compile warnings but i think they're benign.

```plaintext
src/llama.cpp: In function ‘ggml_type img_tensor_get_type(quantize_state_internal&, ggml_type, const ggml_tensor*, llama_ftype)’:
src/llama.cpp:15871:33: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
15871 |         (arch == LLM_ARCH_FLUX) && (
      |         ~~~~~~~~~~~~~~~~~~~~~~~~^~~~
15872 |             (name.find("img_in.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15873 |             (name.find("time_in.in_layer.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15874 |             (name.find("vector_in.in_layer.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15875 |             (name.find("guidance_in.in_layer.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15876 |             (name.find("final_layer.linear.") != std::string::npos)
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15877 |         ) || (arch == LLM_ARCH_SD1 || arch == LLM_ARCH_SDXL) && (
      |         ~                        
src/llama.cpp:15885:33: warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
15885 |         (arch == LLM_ARCH_FLUX) && (
      |         ~~~~~~~~~~~~~~~~~~~~~~~~^~~~
15886 |             (name.find("txt_in.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15887 |             (name.find("time_in.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15888 |             (name.find("vector_in.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15889 |             (name.find("guidance_in.") != std::string::npos) ||
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15890 |             (name.find("final_layer.") != std::string::npos)
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
15891 |         ) || (arch == LLM_ARCH_SD1 || arch == LLM_ARCH_SDXL) && (
      |         ~                        
src/llama.cpp: In function ‘llama_rope_type llama_rope_type(const llama_model*)’:
src/llama.cpp:17589:12: warning: enumeration value ‘LLM_ARCH_FLUX’ not handled in switch [-Wswitch]
17589 |     switch (model->arch) {
      |            ^
src/llama.cpp:17589:12: warning: enumeration value ‘LLM_ARCH_SD1’ not handled in switch [-Wswitch]
src/llama.cpp:17589:12: warning: enumeration value ‘LLM_ARCH_SDXL’ not handled in switch [-Wswitch]
```

looking at the heuristic part, not really sure what it's complaining about. the rope type thing is probably unreachable code for this kind of model.

i synchronized with your changes.

city96: Size slightly increased but I think it's worth it. For me it made it a lot more coherent.

Odd how mine doesn't complain about either. I've removed the rope stuff since we never hit that switch as far as I can tell but didn't get a warning there either.

image

blepping: probably just a different C++ compiler version (i'm on GCC 14.2.1 which is probably more recent than most people). also compiling with `make` rather than `cmake` could be a factor.

not a problem or anything, just thought i'd mention it.

> Size slightly increased but I think it's worth it. For me it made it a lot more coherent.

weirdly i got worse results. :) just the luck of the seed change lottery. or maybe i need to keep my SD models from getting too smart by smashing down the weights.

***

unrelated question that probably shouldn't be here: is there any conceivable way to generate an importance matrix for these models? i have no idea how that would work.

city96: >weirdly i got worse results. :) just the luck of the seed change lottery. 

We can always adjust the weight logic, nothing is set it stone if it turns out to be worse in some way. It's mostly just the embedding keys that I added since without those it was ignoring most of the prompt or falling apart at Q3_K_M lol.

>is there any conceivable way to generate an importance matrix for these models?
>i have no idea how that would work.

That makes two of us :,)
I still have to look into that, probably with T5 first since that's supported by llama.cpp but doesn't work with imatrix either.

city96:
Great work on getting this finalized @blepping + thanks for the help/pointers @compilade!

xueqing0622:
https://github.com/city96/ComfyUI-GGUF/tree/main/tools

blepping:
@xueqing0622 You need to follow the instructions at the URI you linked. It looks like you're using an unpatched version of `llama.cpp`, it requires a patch since `llama.cpp` doesn't have any built-in support for image models.

xueqing0622:
> @xueqing0622 You need to follow the instructions at the URI you linked. It looks like you're using an unpatched version of `llama.cpp`, it requires a patch since `llama.cpp` doesn't have any built-in support for image models.

yes, it works, thx blepping. Have a nice day!

PR #80, opened 2024-08-28
Support loading u-net from stable-diffusion.cpp GGUF files

### LINKED ISSUE #53: support stable-diffusion.cpp gguf models for SD1/SDXL
i made sdxl gguf model ,, unfournutly the node not support it !

blepping:
`stable-diffusion.cpp` quantizes some extra layer types so some more operations needed to be added. the argument parsing for those is sort of brittle, not really sure how to make it better.

they prefix their keys with `model.diffusion_model.` - maybe we should too to enable loading from GGUF checkpoints rather than just u-net, CLIP, etc separately.

some initial work to make `gguf_sd_loader` usable for loading from different key prefixes.

some other small cleanups like using tuple rather than a mutable list type for short constant sequences, avoid recalculating lengths/file lists unnecessarily. couldn't help myself, but if you want i can remove those unrelated changes.

still seems to work with the existing `GGUF` u-net files i tested with as well as output from `stable-diffusion.cpp` which has u-net, CLIP and baked VAE (but we only currently load the u-net keys).

lightly tested but theoretically could be merged in its current state.

may close #53

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -32,27 +32,50 @@ def gguf_sd_loader_get_orig_shape(reader, tensor_name):
         raise TypeError(f"Bad original shape metadata for {field_key}: Expected ARRAY of INT32, got {field.types}")
     return torch.Size(tuple(int(field.parts[part_idx][0]) for part_idx in field.data))
 
-def gguf_sd_loader(path):
+def gguf_sd_loader(path, handle_prefix="model.diffusion_model.", require_prefix=False):
     """
     Read state dict as fake tensors
     """
     reader = gguf.GGUFReader(path)
     sd = {}
     dt = {}
+    arch_field = reader.get_field("general.architecture")
+    if arch_field is not None:
+        if len(arch_field.types) != 1 or arch_field.types[0] != gguf.GGUFValueType.STRING:
+            raise TypeError(f"Bad type for GGUF general.architecture key: expected string, got {arch_field.types!r}")
+        arch_str = str(arch_field.parts[arch_field.data[-1]], encoding="utf-8")
+        if arch_str not in {"flux", "sd1", "sdxl"}:
+            raise ValueError(f"Unexpected architecture type in GGUF file, expected one of flux, sd1, sdxl but got {arch_str!r}")
+    else:
+        arch_str = None
+    if handle_prefix is not None:
+        prefix_len = len(handle_prefix)
+        tensor_names = set(tensor.name for tensor in reader.tensors)
+        has_prefix = any(s.startswith(handle_prefix) for s in tensor_names)
+    else:
+        has_prefix = False
+    if require_prefix and not has_prefix:
+        raise RuntimeError(f"Loader requires prefix {handle_prefix!r} which does not exist in the model")
     for tensor in reader.tensors:
-        tensor_name = str(tensor.name)
+        sd_key = tensor_name = tensor.name
+        if has_prefix:
+            if not tensor_name.startswith(handle_prefix):
+                continue
+            sd_key = tensor_name[prefix_len:]
+        tensor_type_str = str(tensor.tensor_type)
         torch_tensor = torch.from_numpy(tensor.data) # mmap
         shape = gguf_sd_loader_get_orig_shape(reader, tensor_name)
         if shape is None:
             shape = torch.Size(tuple(int(v) for v in reversed(tensor.shape)))
-        elif tensor.tensor_type in {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}:
+        if tensor.tensor_type in {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}:
             torch_tensor = torch_tensor.view(*shape)
-        sd[tensor_name] = GGMLTensor(
+        sd[sd_key] = GGMLTensor(
             torch_tensor,
             tensor_type = tensor.tensor_type,
-            tensor_shape = shape
+            tensor_shape = shape,
+            ggml_tensor_name = tensor_name,
         )
-        dt[str(tensor.tensor_type)] = dt.get(str(tensor.tensor_type), 0) + 1
+        dt[tensor_type_str] = dt.get(tensor_type_str, 0) + 1
 
     # sanity check debug print
     print("\nggml_sd_loader:")
@@ -172,14 +195,14 @@ class UnetLoaderGGUF:
     def load_unet(self, unet_name, dequant_dtype=None, patch_dtype=None, patch_on_device=None):
         ops = GGMLOps()
 
-        if dequant_dtype in ["default", None]:
+        if dequant_dtype in ("default", None):
             ops.Linear.dequant_dtype = None
         elif dequant_dtype in ["target"]:
             ops.Linear.dequant_dtype = dequant_dtype
         else:
             ops.Linear.dequant_dtype = getattr(torch, dequant_dtype)
 
-        if patch_dtype in ["default", None]:
+        if patch_dtype in ("default", None):
             ops.Linear.patch_dtype = None
         elif patch_dtype in ["target"]:
             ops.Linear.patch_dtype = patch_dtype
@@ -283,11 +306,12 @@ class CLIPLoaderGGUF:
 class DualCLIPLoaderGGUF(CLIPLoaderGGUF):
     @classmethod
     def INPUT_TYPES(s):
+        file_options = (s.get_filename_list(), )
         return {
             "required": {
-                "clip_name1": (s.get_filename_list(), ),
-                "clip_name2": (s.get_filename_list(), ),
-                "type": (["sdxl", "sd3", "flux"], ),
+                "clip_name1": file_options,
+                "clip_name2": file_options,
+                "type": (("sdxl", "sd3", "flux"), ),
             }
         }
 
@@ -296,18 +320,19 @@ class DualCLIPLoaderGGUF(CLIPLoaderGGUF):
     def load_clip(self, clip_name1, clip_name2, type):
         clip_path1 = folder_paths.get_full_path("clip", clip_name1)
         clip_path2 = folder_paths.get_full_path("clip", clip_name2)
-        clip_paths = [clip_path1, clip_path2]
+        clip_paths = (clip_path1, clip_path2)
         clip_type = clip_name_dict.get(type, comfy.sd.CLIPType.STABLE_DIFFUSION)
         return (self.load_patcher(clip_paths, clip_type, self.load_data(clip_paths)),)
 
 class TripleCLIPLoaderGGUF(CLIPLoaderGGUF):
     @classmethod
     def INPUT_TYPES(s):
+        file_options = (s.get_filename_list(), )
         return {
             "required": {
-                "clip_name1": (s.get_filename_list(), ),
-                "clip_name2": (s.get_filename_list(), ),
-                "clip_name3": (s.get_filename_list(), ),
+                "clip_name1": file_options,
+                "clip_name2": file_options,
+                "clip_name3": file_options,
             }
         }
 
@@ -317,7 +342,7 @@ class TripleCLIPLoaderGGUF(CLIPLoaderGGUF):
         clip_path1 = folder_paths.get_full_path("clip", clip_name1)
         clip_path2 = folder_paths.get_full_path("clip", clip_name2)
         clip_path3 = folder_paths.get_full_path("clip", clip_name3)
-        clip_paths = [clip_path1, clip_path2, clip_path3]
+        clip_paths = (clip_path1, clip_path2, clip_path3)
         clip_type = clip_name_dict.get(type, comfy.sd.CLIPType.STABLE_DIFFUSION)
         return (self.load_patcher(clip_paths, clip_type, self.load_data(clip_paths)),)
 
--- a/nodes.py
+++ b/nodes.py
@@ -32,7 +32,7 @@ def gguf_sd_loader_get_orig_shape(reader, tensor_name):
         raise TypeError(f"Bad original shape metadata for {field_key}: Expected ARRAY of INT32, got {field.types}")
     return torch.Size(tuple(int(field.parts[part_idx][0]) for part_idx in field.data))
 
-def gguf_sd_loader(path, handle_prefix="model.diffusion_model.", require_prefix=False):
+def gguf_sd_loader(path, handle_prefix="model.diffusion_model."):
     """
     Read state dict as fake tensors
     """
@@ -54,8 +54,6 @@ def gguf_sd_loader(path, handle_prefix="model.diffusion_model.", require_prefix=
         has_prefix = any(s.startswith(handle_prefix) for s in tensor_names)
     else:
         has_prefix = False
-    if require_prefix and not has_prefix:
-        raise RuntimeError(f"Loader requires prefix {handle_prefix!r} which does not exist in the model")
     for tensor in reader.tensors:
         sd_key = tensor_name = tensor.name
         if has_prefix:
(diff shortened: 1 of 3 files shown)


blepping:
actually, SDXL doesn't work - fails on missing key `conv_in.weight`. that doesn't exist in our GGUF file either but still works, so i think something in model detection is making it take an alternative code path. not sure exactly what's going wrong.

ops.py
```diff
@@ -9,20 +9,22 @@ class GGMLTensor(torch.Tensor):
     """
     Main tensor-like class for storing quantized weights
     """
-    def __init__(self, *args, tensor_type, tensor_shape, patches=[], **kwargs):
+    def __init__(self, *args, tensor_type, tensor_shape, patches=[], ggml_tensor_name=None, **kwargs):
```
city96, line 12: Why are we storing tensor_name as an attribute? This should be handled by the state dict and there's no reason it should ever be part of the tensor-like object.

nodes.py
```diff
@@ -32,27 +32,50 @@ def gguf_sd_loader_get_orig_shape(reader, tensor_name):
...
+    else:
+        arch_str = None
+    if handle_prefix is not None:
+        prefix_len = len(handle_prefix)
+        tensor_names = set(tensor.name for tensor in reader.tensors)
+        has_prefix = any(s.startswith(handle_prefix) for s in tensor_names)
+    else:
+        has_prefix = False
+    if require_prefix and not has_prefix:
+        raise RuntimeError(f"Loader requires prefix {handle_prefix!r} which does not exist in the model")
```
city96, line 58: I don't think we need "require_prefix" as I'm vehemently against ever supporting multiple models in a single GGUF file as that's 100% against what the format specs intend.

I think we can just use this same bit of code to strip out the unet with the prefix hardcoded as "model.diffusion_model." since [even SD3 uses that same prefix](https://huggingface.co/stabilityai/stable-diffusion-3-medium/tree/main?show_file_info=sd3_medium.safetensors) - we can turn it into a loop and add more if that ever changes.

https://github.com/city96/ComfyUI-GGUF/blob/3d9869240bd4ff786030ad85b09441045bb3fda9/tools/convert.py#L62-L70

city96 (changes requested):

city96:
The shape of that key is used to [detect the model channels](https://github.com/comfyanonymous/ComfyUI/blob/38c22e631ad090a4841e4a0f015a30c565a9f7fc/comfy/model_detection.py#L372), so it could be that. It's probably returning the shape of the reshaped tensor causing it to fail. Are they storing the original tensor shape anywhere in the file?

ops.py
blepping: i can remove it if you want. my thinking was it could make debugging a lot easier since currently if you have issues with a `GGMLTensor` there's really no way to tell where it came from.

nodes.py
blepping: > I don't think we need "require_prefix" as I'm vehemently against ever supporting multiple models in a single GGUF file as that's 100% against what the format specs intend.

i can remove it if you insist but i _really_ hope you'll reconsider. combined checkpoints are way, way easier to deal with from a user perspective. splitting your existing checkpoints into separate VAE, CLIP and u-net and then having to quantize them individually is both a hassle to do and harder to manage even after they've been quantized.

also the `stable-diffusion.cpp` GGUF file already has VAE, CLIP and u-net combined: you can't make them stop doing that. the most you could do by removing support here is make it some the user is only able to use the u-net part and the rest is just useless to them. they'll have to do unnecessary work to get VAE/CLIP that already exists in the GGUF file.

blepping:
> The shape of that key is used to [detect the model channels](https://github.com/comfyanonymous/ComfyUI/blob/38c22e631ad090a4841e4a0f015a30c565a9f7fc/comfy/model_detection.py#L372), so it could be that.

actually found that stuff - turns out their GGUF file has the shape padded with 1 dimensions, so ours would be `[640, 640]` and theirs is `[640, 640, 1, 1]`. can we just squeeze the last dimension however many times when loading?

nodes.py
```diff
@@ -32,27 +32,55 @@ def gguf_sd_loader_get_orig_shape(reader, tensor_name):
...
         torch_tensor = torch.from_numpy(tensor.data) # mmap
         shape = gguf_sd_loader_get_orig_shape(reader, tensor_name)
         if shape is None:
             shape = torch.Size(tuple(int(v) for v in reversed(tensor.shape)))
-        elif tensor.tensor_type in {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}:
+            if arch_str is None and (tensor_name.endswith(".proj_in.weight") or tensor_name.endswith(".proj_out.weight")):
+                # Workaround for stable-diffusion.cpp SDXL detection.
+                # Impossible to land here for our own GGUF files as we set architecture.
+                while len(shape) > 2 and shape[-1] == 1:
+                    shape = shape[:-1]
```
blepping, line 74: this isn't amazingly elegant but it will only have an effect when loading `stable-diffusion.cpp` GGUF files. they don't store an architecture so there's no way to know exactly what type of model we're dealing with without adding heuristics.

if we are ever loading a non-SDXL `stable-diffusion.cpp` format GGUF file which is incompatible squeezing those dimensions that then this will have to be reevaluated.

ops.py
city96: Fair enough, think you should remove it in the final version, it's just another thing to be copied around and if by chance the tensor attributes get stripped then this one will just be `None` as well.

nodes.py
city96: Looks good enough to me as a workaround for now. As you say, we can always deal with it later if it comes to it.

nodes.py
city96: Pretty sure almost all models use the same 4 VAE models and the same 3-4 CLIP encoders anyway. Is there really any point in copying these? Besides, we'd then have to deal with the VAE model (which is like 300MB and probably shouldn't be quantized anyway, since most of it's memory usage is during runtime).

Then you'd have the issue about people being confused if a given gguf file is the actual unet or a combined checkpoint one. "which loader do I use" etc. With llama.cpp, even the mmproj model is split as a separate file for vllms.

Besides, this repo remains to mostly be centered around large models like flux and the upcoming sd3.1 (assuming it'll be larger than sd3). With those, packaging the T5 model would just result in comically large checkpoints. And I already know people would start packaging them and releasing crap like "flux1-dev-schnell-merge_q4_k_s_with_T5_q5_k_m.gguf" given the chance lol.

image

blepping: okay, i removed `require_prefix`. stuff like custom VAEs (and trained TE) is more common for SD15 models. i think it would be good from a user perspective if they could treat `.gguf` files similarly to `.safetensors` but it's your project.

leaving the prefix code relatively generic rather than hardcoding it doesn't increase the complexity much: i left that as is. if you ever do want to support different prefixes later on it will be easier to add, otherwise you can just never pass that parameter and it be effectively the same as hardcoding `model.diffusion_model.`

city96: Yeah sure that's fine, we'll probably have to move that loader function anyway assuming I ever get the time to work on the c++ stuff.

city96 (approved):

city96:
This works for SD but for Flux the group norm fails with the following:

image

Those tensors should be small enough so maybe we can do something like what's done with the text encoder and just dequantize them on load? I think the less layers we mess with the better in this case, and conv2d/linear are the two main ones that make up the bulk of the data anyway. That has less of a chance of breaking things like clip/controlnet/etc (which may already be broken since they can't grab the `layer.in_features` attribute).

https://github.com/city96/ComfyUI-GGUF/blob/abe5c6a6c90a3f8a53a097614bf749de506c6df7/nodes.py#L93-L96

Or I guess you could refactor and inherit from `nn.[whatever]`, do the `super().init`, delete the weights, then reuse all the `.self` attributes. You'd replace the forward  with `*args, **kwargs` and pass them to the other one, though that might just be even more annoying and hacky lol.

blepping:
> This works for SD but for Flux the group norm fails with the following

do you mean these changes break the previously working ComfyUI-GGUF Flux or that Flux created with `stable-diffusion.cpp` is all that fails?

city96:
Previously working ones created by our own conversion stuff.

nodes.py
blepping: there are GGML bindings for Python. if it was me, what i'd probably do is look to make my own conversion/quantization script and completely avoid having to mess with llama.cpp, patch GGML tensor name length, etc. that would be somewhat more up front effort but maintaining patches is basically unbounded effort.

blepping:
> [Previously working ones](https://huggingface.co/city96/FLUX.1-dev-gguf) created by our own conversion stuff.

that's unfortunate.

> Those tensors should be small enough so maybe we can do something like what's done with the text encoder and just dequantize them on load? 

are you saying there's a way to know what type of operation a tensor is going to be used for at load time? the first thing we could do if that's the case is only enable ops if there's at least one quantized tensor that uses it. `stable-diffusion.cpp`-quantized Flux models still may not work but that would at least eliminate the possibility of breaking the GGUF files `ComfyUI-GGUF` created itself.

blepping:
@city96 just checking to make sure you didn't miss the above question amidst other notifications. i'll probably have time to work on this early tomorrow, right now it's not clear exactly what to do though. if there is a relatively easy way to know what operation the tensors will be used for at the very least i can make sure it doesn't break any of the existing stuff. would also need to know how to do that to take your advice about just pre-dequantizing those tensors also.

if you just didn't get a chance to respond yet or are still thinking about it, not pressing you or anything.

city96:
>just checking to make sure you didn't miss the above question amidst other notifications.

I did in fact miss it lol. I had a couple ideas on how to deactivate ops, one was some logic based around this part as we hijack it anyway:

https://github.com/city96/ComfyUI-GGUF/blob/abe5c6a6c90a3f8a53a097614bf749de506c6df7/ops.py#L87-L93

We could possibly do a switch where the layer is initialized normally, and if neither the bias nor the weight are quantized, then they're just loaded as regular weights and the original ComfyUI forward is used for the ops. If they are quantized, then we delete/overwrite the weight with the quantized one. Not the best but it's an idea. Beats guessing based on the tensor names I guess.

Though what you say is definitely what we should be doing:

>the first thing we could do if that's the case is only enable ops if there's at least one quantized tensor that uses it.

i.e. no reason to ever touch conv2d for flux, or linear when someone is loading a regular clip model using the gguf loader/etc.

city96:
@blepping okay, [check this out](https://github.com/city96/ComfyUI-GGUF/tree/ops_refactor). I've refactored around the original comfy stuff. Somehow it's also way faster in lowvram mode.

Anyway, you should be able to add more layers by pasting the `forward_comfy_cast_weights` function from `ComfyUI/comfy/ops.py`

blepping:
sorry, got distracted by other stuff. hopefully i will get a chance to work on this tomorrow and finish it up. (no response necessary)

ops.py
```diff
@@ -59,17 +59,39 @@ class GGMLLayer(torch.nn.Module):
     comfy_cast_weights = True
     dequant_dtype = None
     patch_dtype = None
-
-    def _load_from_state_dict(self, state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs):
+    ggml_delegate_types = {None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}
+
+    def ggml_check_delegate(self, *, weight=None, bias=None):
```
blepping, line 64: @city96 what do you think of this approach?

we can delegate both the private Torch methods (scary to override) and op functions in `manual_cast` whenever there are no quantized tensors to deal with and hopefully make breaking stuff less likely.

could also potentially just add a handler for all the ops types without making non-quantized stuff break.

blepping:
okay, current iteration is tested to work with:

* t5xxl (from your HF repo)
* SD 1.5 (both yours and stable-diffusion.cpp)
* SDXL (both yours and stable-diffusion.cpp)
* Flux (both yours and stable-diffusion.cpp)

sorry, i should have tested with Flux and T5 before instead of making you do the work.

**note**: i changed this pull to point at your `ops_rework` branch. just mentioning it since if you do something like merge `ops_rework` and _then_ merge this into `ops_rework` it could be awkward.

ops.py
city96: Not bad, I think that should work. One small nitpick is that maybe `ggml_check_delegate` could be `is_quantized` instead with the logic flipped since that makes it a lot more apparent what this does. When reading over it I kept going "wait why is- oh nvm delegate means the regular ones" lmao. (also, better handle both args and kwargs for the input).

```py
    def is_quantized(self, *args, weight=None, bias=None, **kwargs):
        if weight is None:
            weight = self.weight
        if bias is None:
            bias = self.bias
        weight_type = getattr(weight, "tensor_type", None)
        bias_type = getattr(bias, "tensor_type", None) if bias is not None else None
        torch_compatible = {None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}
        return weight_type not in torch_compatible and bias_type not in torch_compatible
```

blepping: no problem about changing the name/flipping the logic. what are you doing with `*args` and `**kwargs` though? it only optionally gets passed specific weight/bias to use or nothing, it's not the actual wrapper - just a function that returns a boolean.

blepping: i renamed the function to `is_ggml_quantized` and reversed the logic. i didn't bring the list of Torch compatible types into the function because there's no need to rebuild that set every single time the function is called.

the thinking behind calling it `is_ggml_quantized` rather than `is_quantized` is that we're subclassing both Torch modules and Comfy stuff and i want to avoid clobbering something that already exists. so i tried to choose names that aren't likely to conflict with any existing method/variable/etc.

of course, if you want to name it exactly something that's no problem and i can change it to whatever you want.

city96: Isn't it technically still like `*args`? since it just "consumes" any positional arguments but without assigning it to a value? I could be wrong (it also doesn't matter in this case because yeah, as you say it won't be used for anything else).

Also, what do you think about moving some of this stuff to dequant.py in the form of a quick check like this? Since we have to check this in other places like nodes.py anyway. Was thinking something like this:

image

blepping: > Isn't it technically still like *args? since it just "consumes" any positional arguments but without assigning it to a value? 

nope, it actually doesn't consume any arguments. it's just a marker that any following arguments must be passed as keyword arguments.

```plaintext
>>> def testy(a, *, x=None, y=None): print(a, x, y)
... 
>>> testy(1)
1 None None
>>> testy(1, x=2, y=3)
1 2 3
>>> testy(1, 2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: testy() takes 1 positional argument but 2 were given
```

you can also define a function like `def(blah, *, x, y): ...` and `x` and `y` _still_ have to be passed as keyword arguments, however since they have no default value they aren't optional: you must pass them or hit an exception.

***

> Also, what do you think about moving some of this stuff to dequant.py in the form of a quick check like this?

sure, that sounds reasonable. i think you're right that it probably makes more sense there than as a method.

city96: Huh. Learned something new today. You can tell all I ever used python for was random automation crap.

(Also, don't think we need to add the `_ggml_` into function names that often unless it's something that already has that. So for the load state dict it makes sense but maybe not as much for `is_quantized`? Though most of the naming in the repo is pretty random anyway)

blepping: if left `ggml` in the method name (couldn't easily be eliminated) but refactored it based on your example. how about this iteration?

city96: Yeah the new changes look good, thank you.

The merge conflict is on me, though I think at least it fixed the damn double RAM/random OOM issue. LMK if you think it's ready to merge into the `ops_refactor` branch. I'll ask for a review from you when we merge that one to master to make sure I didn't break anything.

blepping: i resolved the conflicts, still seems to work with Flux at least. i have to go for now, i won't get a chance to test other models until tomorrow unfortunately. not sure if you want to wait or not.

i've been testing with all but your most recent changes

city96: Sounds good, I'll merge it to this branch and do any changes on my end (i.e. switching away from mmap loading since it's broken on latest comfy anyway and doesn't save any time/memory).

We're not in any hurry so I'll just wait for your final OK before doing the merge with main whenever you're free.

PR #92, opened 2024-09-02
Fix LoRA reloading, ops refactor, better lowvram speed, sd.cpp compat

city96:
Changes:
- LoRAs no longer take 2x system RAM for no reason
- LoRA/etc should no longer reload the model on weight changes or when enabling/disabling/muting them
- LowVRAM mode shouldn't completely destroy inference speed anymore. [Q8_0 on 10GB from 3.49 s/it to 2.35 s/it @ 1024]
- ops now matches the ComfyUI equivalent more closely
- ops no longer triggers for FP16/FP32 weights (by @blepping )
- SDXL models made with stable-diffusion.cpp can now be loaded (by @blepping )

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,13 +14,13 @@ Note: The "Force/Set CLIP Device" is **NOT** part of this node pack. Do not inst
 > [!IMPORTANT]  
 > Make sure your ComfyUI is on a recent-enough version to support custom ops when loading the UNET-only.
 
-To install the custom node normally, git clone this repository and install the only dependency for inference (`pip install --upgrade gguf`)
+To install the custom node normally, git clone this repository into your custom nodes folder (`ComfyUI/custom_nodes`) and install the only dependency for inference (`pip install --upgrade gguf`)
 
 ```
 git clone https://github.com/city96/ComfyUI-GGUF
 ```
 
-To install the custom node on standalone, open a CMD inside the "ComfyUI_windows_portable" folder (where your `run_nvidia_gpu.bat` file is) and use the following commands:
+To install the custom node on a standalone ComfyUI release, open a CMD inside the "ComfyUI_windows_portable" folder (where your `run_nvidia_gpu.bat` file is) and use the following commands:
 
 ```
 git clone https://github.com/city96/ComfyUI-GGUF ComfyUI/custom_nodes/ComfyUI-GGUF
--- a/dequant.py
+++ b/dequant.py
@@ -45,6 +45,10 @@ def split_block_dims(blocks, *args):
     dims = list(args) + [n_max - sum(args)]
     return torch.split(blocks, dims, dim=1)
 
+# Full weights #
+def dequantize_blocks_BF16(blocks, block_size, type_size, dtype=None):
+    return (blocks.view(torch.int16).to(torch.int32) << 16).view(torch.float32)
+
 # Legacy Quants #
 def dequantize_blocks_Q8_0(blocks, block_size, type_size, dtype=None):
     d, x = split_block_dims(blocks, 2)
@@ -222,6 +226,7 @@ def dequantize_blocks_Q2_K(blocks, block_size, type_size, dtype=None):
     return qs.reshape((n_blocks, -1))
 
 dequantize_functions = {
+    gguf.GGMLQuantizationType.BF16: dequantize_blocks_BF16,
     gguf.GGMLQuantizationType.Q8_0: dequantize_blocks_Q8_0,
     gguf.GGMLQuantizationType.Q5_1: dequantize_blocks_Q5_1,
     gguf.GGMLQuantizationType.Q5_0: dequantize_blocks_Q5_0,
--- a/dequant.py
+++ b/dequant.py
@@ -1,24 +1,18 @@
 # (c) City96 || Apache-2.0 (apache.org/licenses/LICENSE-2.0)
 import gguf
 import torch
-import numpy as np
 from tqdm import tqdm
 
-def dequantize_tensor(tensor, dtype=torch.float16):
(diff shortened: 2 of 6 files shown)


city96:
Tagging @blepping for a final sanity check before merging.

I tried to clean up the part with the model loading + move the import on the off chance I'm not the only one who messed around with the stuff in `tools` (the import + model detection will only run when there's no architecture string in the GGUF file now)

blepping:
@city96 tested with all the permutations of sd.cpp and models converted with the convert script here: sdxl, sd15, Flux, t5xxl (which i guess was converted by llama.cpp technically).

i only tested a 1-2 quant formats, and my test was just to try to generate something and make sure it loaded/produced a reasonable result. so we could say that's not very scientific or exhaustive.

anyway, for what it's worth, i didn't find any issues!

the only thing i'd mention (and i might just be mistaken, since it's very subjective) is that after syncing with the ops refactor stuff the other day it seems like there's a fairly long pause before the first sampler step that didn't occur before. i would _guess_ it's due to not using mmap now, not sure if there's an easy way to toggle that to double check it. i'm also on Linux so it's possible mmap stuff had issues on Windows before but not Linux?

i wouldn't consider that a blocker (and it seems like less of a factor after the disk cache is hot) but might be something to look at later on.

city96:
Yeah, the mmap thing is definitely annoying to deal with. After the first load it does technically unload but if we completely disable it we're back at the 2x system RAM spike (currently it loads whatever it can onto the GPU, rest into system RAM, then the mmap gets unloaded). I haven't found any way to unload parts of the mmap thing on windows individually.

blepping:
what's actually controlling whether it uses mmap or not? there aren't any explicit references to `mmap` stuff, so i guess the difference is something like a copying out of the original tensor so the reference can get dropped.

>  we're back at the 2x system RAM spike 

are you sure it's actually using 2x system RAM? if one half you're talking about is the OS disk cache, that basically is the equivalent of free memory: it'll just get discarded if the memory is needed.

also, depending on where the choice of using mmap or not is, it might be possible to use mmap except when patches are defined.

city96:
Windows doesn't over-commit memory, which is the main reason it's causing issues, same thing with safetensors iirc. gguf-py uses mmap by default, you can check the underlying object + it's the source of this warning:

```
UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at ..\torch\csrc\utils\tensor_numpy.cpp:212.)
  torch_tensor = torch.from_numpy(tensor.data) # mmap
```

Once we copy it into a parameter and move it to any device it turns into a normal tensor (I think) so it should be dropped after the model is first loaded to the GPU. You can force unload the underlying mmap object but this just hard crashes it if you ever accidentally access it again.

blepping:
> Windows doesn't over-commit memory, which is the main reason it's causing issues

is there any relatively simple way to leave it enabled for non-Windows users? if you point me in the right direction, i might be able to implement that.

city96:
Not that I know of other than the thing we had before where we used the tensors directly instead of assigning them to parameters but that's pretty fragile.

Also need to start looking at inference speed, I swear it dropped compared to the original version of this repo on old versions of comfy lol.

city96:
Looks like the model management does expect `unpatch_model` to move the model off the device after all, so it'll probably need a different fix to stop the patches reloading after all.

I pushed a hotfix to respect disable smart memory (i.e. always unload no matter what) but lowvram is off because it seems to lose track how how much memory the model uses after it gets unloaded or unpatched once.

city96:
@RandomGitUser321 
Yeah, this'll definitely need either a different workaround or coordination with the actual ComfyUI repo for a proper solution. The main issue is that patching/unpatching and loading/unloading are too closely coupled.

Could you make an issue about it? Just so people can tell that lowvram + loading/unloading is borked lol.

@blepping possible pointers for mmap:

Changing this line to have a `.clone()` at the end cuts inference RAM usage in half when lowvram is used, meaning mmap isn't being unloaded. I think this is due to `.to()` making a copy of the tensor (when moved to the GPU, i.e. no longer tied to mmap) but the lowvram ones never get moved anywhere so they just stay linked to mmap forever, meaning it won't get closed/gc'd.

https://github.com/city96/ComfyUI-GGUF/blob/6dbb4ba989d344a31a7d74eed983475169fa6b27/nodes.py#L79

Only workaround I can think of is making a copy of all CPU weights somehow, maybe after `load` in the model patcher but that's, yeah, no.

blepping:
> Changing this line to have a `.clone()` at the end cuts inference RAM usage in half when lowvram is used, meaning mmap isn't being unloaded.

are you saying that the mmap isn't being unloaded _without_ that change or making that change causes it not to be unloaded?

city96:
@blepping here's what I'm seeing:
- No lowvram -> whole model gets moved to GPU -> weights no longer linked to mmap -> all good
- lowvram but adding .copy() -> weights are copied and no longer linked to mmap -> all good (but uses more memory during load)
- lowvram with no extra changes -> some weights stay linked to mmap -> wastes 2x memory

city96:
Possible fix for the system RAM issue by making sure CPU weights are moved to GPU at least once (can probably find a better way for this) https://github.com/city96/ComfyUI-GGUF/commit/c8923a4cd7b7ddea104ea1939b4d715bf1849ccb

Doesn't change model management stuff.

city96:
Disabled the instant patch logic for now, will need a better fix to that https://github.com/city96/ComfyUI-GGUF/commit/717a0e1d0b321b0ae8243e26328d6bb55ceac288

PR #154, opened 2024-11-05
Add mochi to CLIPLoaderGGUF options

RandomGitUser321:
ComfyUI implemented it today, so this should be all that's needed to use the T5 GGUF with it. I'm testing it right now, but it's already diffusing(slow on my GPU at a minute per step). I'd assume that it would have spit some error by now if this PR didn't work.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -295,7 +295,7 @@ class CLIPLoaderGGUF:
         return {
             "required": {
                 "clip_name": (s.get_filename_list(),),
-                "type": (["stable_diffusion", "stable_cascade", "sd3", "stable_audio"],),
+                "type": (["stable_diffusion", "stable_cascade", "sd3", "stable_audio", "mochi"],),
             }
         }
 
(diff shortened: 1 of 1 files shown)


city96:
You need to add it to `clip_name_dict` above for it to be correct I think? Otherwise it just defaults to God knows what lol.

```py
    "mochi": comfy.sd.CLIPType.MOCHI,
```

But then that'd break older installs, so you'd need to add an if or use a `getattr`:

```py
    "mochi": getattr(comfy.sd.CLIPType, "MOCHI", None),
```

(I have absolutely no clue if this works I'm not at my main PC)

city96:
Neat, thanks. Looks like the logic in comfy [defaults to mochi if not set to SD3](https://github.com/comfyanonymous/ComfyUI/blob/8afb97cd3f4d74d3a09fb2c3bd89488fd855d2ad/comfy/sd.py#L507-L513), probably why it still worked while set to SD. Still, good to have it pass the correct value for future models and stuff.

PR #216, opened 2025-02-20
convert.py/patch refactor

city96:
This currently adds:
- The updated patch that has ltxv, wan2.1 and hunyuan key names (replaces the special SD3 patch to make it less confusing)
- The extra script required for the janky 4 step process required by hunyuan and wan + the readme for it
- Some commits from a month ago that reduce the number of tensors kept in FP32 + add sanity checks
- Some minimal refactoring on the convert script

Not 100% confident in merging yet, will add other model support here & do a full sweep of the other models before this can be merged.

Possible TODOs/ideas:
- Use comfy native model detection logic for convert script
- ~~clean up fix_5d_tensors.py~~ done


city96:
Tagging @kijai and @pollockjj who requested the updated code - this is the same as what I shared earlier but you guys can follow the PR for updates.

city96:
Merging for now as this needs to be in main for compat model loading as well as for wip all in one quant script.

PR #243, opened 2025-03-28
Enable more `torch.compile` optimization by reducing graph breaks

StrongerXi:
As title, there were 2 fundamental sources of graph breaks:
1. using a pre-existing set in a `torch.compile` region, this is a [known issue](https://github.com/pytorch/pytorch/issues/145761) in PyTorch and the fix is WIP, but it's a lot easier to change the set into a tuple here to fix the graph break.
2. using tensor subclass `GGMLTensor`. The full support of tensor subclass has been gated for `torch.compile`, but with [a recent WIP stack](https://github.com/pytorch/pytorch/pull/149792), we'll be able to remove the gating and have `torch.compile` fully support the `GGMLTensor` in this repo. The only required change here is avoid changing the `__class__` attribute, and instead using `torch.Tensor` to convert the subclass object back to Tensor.

### DIFF
--- a/ops.py
+++ b/ops.py
@@ -6,14 +6,34 @@ import comfy.ops
 import comfy.model_management
 from .dequant import dequantize_tensor, is_quantized
 
-# to avoid breaking really old pytorch versions
-if hasattr(torch, "compiler") and hasattr(torch.compiler, "disable"):
-    torch_compiler_disable = torch.compiler.disable
-else:
-    def torch_compiler_disable(*args, **kwargs):
+def chained_hasattr(obj, chained_attr):
+    probe = obj
+    for attr in chained_attr.split('.'):
+        if hasattr(probe, attr):
+            probe = getattr(probe, attr)
+        else:
+            return False
+    return True
+
+# A bakcward and forward compatible way to get `torch.compiler.disable`.
+def get_torch_compiler_disable_decorator():
+    def dummy_decorator(*args, **kwargs):
         def noop(x):
             return x
         return noop
+
+    from packaging import version
+
+    if not chained_hasattr(torch, "compiler.disable"):
+        return dummy_decorator # torch too old
+    elif version.parse(torch.__version__) >= version.parse("2.8"):
+        return dummy_decorator # torch compile works
+    if chained_hasattr(torch, "_dynamo.config.nontraceable_tensor_subclasses"):
+        return dummy_decorator # torch compile works, nightly before 2.8 release
+    else:
+        return torch.compiler.disable
+
+torch_compiler_disable = get_torch_compiler_disable_decorator()
 
 class GGMLTensor(torch.Tensor):
     """
@@ -149,7 +169,7 @@ class GGMLLayer(torch.nn.Module):
 
         # prevent propagating custom tensor class
         if isinstance(weight, GGMLTensor):
-            weight.__class__ = torch.Tensor
+            weight = torch.Tensor(weight)
 
         # apply patches
         if patch_list:
@@ -189,7 +209,7 @@ class GGMLLayer(torch.nn.Module):
 
         # non-ggml forward might still propagate custom tensor class
         if isinstance(out, GGMLTensor):
-            out.__class__ = torch.Tensor
+            out = torch.Tensor(out)
         return out
 
     def forward_ggml_cast_weights(self, input):
--- a/dequant.py
+++ b/dequant.py
@@ -4,7 +4,7 @@ import torch
 from tqdm import tqdm
 
 
-TORCH_COMPATIBLE_QTYPES = {None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}
+TORCH_COMPATIBLE_QTYPES = (None, gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16)
 
 def is_torch_compatible(tensor):
     return tensor is None or getattr(tensor, "tensor_type", None) in TORCH_COMPATIBLE_QTYPES


ops.py
```diff
@@ -6,8 +6,10 @@
 import comfy.model_management
 from .dequant import dequantize_tensor, is_quantized
 
-# to avoid breaking really old pytorch versions
-if hasattr(torch, "compiler") and hasattr(torch.compiler, "disable"):
+# to avoid breaking really old pytorch version, and avoid graph-breaking really
+# new pytorch version. TODO figure it out, do we have to wait till next release?
+before_torch_gguf_tensor_subclass_support = False
+if hasattr(torch, "compiler") and hasattr(torch.compiler, "disable") and before_torch_gguf_tensor_subclass_support:
```
StrongerXi, line 12: I'm not sure what's the best workaround here... So once https://github.com/pytorch/pytorch/pull/149792 goes in, one can enjoy speeds up by
1. installing torch nightly, and
2. applying the changes in this PR (ideally just pulling ComfyUI-GGUF if we merge this).

But we still need `torch.compiler.disable` so that things don't break for folks without torch nightly. So maybe we need to wait till the next release and check version here, or provide some flag for users with torch nightly to opt-in.

StrongerXi:
I benchmarked this on a single A100 80G, with https://github.com/pytorch/pytorch/pull/149792, on two models:
- [flux1-dev-Q8_0.gguf](https://huggingface.co/city96/FLUX.1-dev-gguf/blob/main/flux1-dev-Q8_0.gguf)
- [wan2.1-t2v-14b-Q4_0.gguf](https://huggingface.co/city96/Wan2.1-T2V-14B-gguf/blob/main/wan2.1-t2v-14b-Q4_0.gguf)

Single batch, did 1 warmup run (which includes compile), and took average of logged prompt execution time from 5 runs (variance is negligible), here are the results, notice
1. the same results for `(main) eager` and `(fixed) eager`
2. speed up from `(fixed) compile` to eager and `(main) compile`.

```verbatim
                | (main) eager | (main) compile | (fixed) eager | (fixed) compile
---------------------------------------------------------------------------------
flux (20 iters) |    12.97s    |     10.95s     |     12.99s    |      8.27s
---------------------------------------------------------------------------------
 wan (30 iters) |   215.89s    |    181.27s     |    215.91s    |    174.48s
```

StrongerXi:
cc @city96

ops.py
city96: Ah, thank you for the follow up on this!
I think we can parse the version number & rework this check. That'd also make a 3 way switch easy since we can just check if the version is [2.1.X or older] OR [nightly after merge date/newer than text stable] to disable the decorator.
I'll see if I can come up with a decent check & report back.

city96:
Nice, looks like a pretty substantial speedup overall. As mentioned in the comment above, we can rework the check that triggers disabling torch compile and once we have a known nightly version that includes the fix we can target it & merge to main.

Ph0rk0z:
We need bleeding edge pytorch for this?

ops.py
city96: Okay, haven't tested it very well, but here is a proposed solution.
Date & version may need to be adjusted depending on what date the new patch is added.

```py
import re
import torch

def get_tc_decorator():
    def dummy_decorator(*args, **kwargs):
        def noop(x):
            return x
        return noop

    try:
        # 3 version numbers, optional date for nightly
        ex = "^(\d+)\.(\d+)\.(\d+)(?:\.dev(\d{8}))?\+.*$"
        m = re.match(ex, torch.version.__version__)
        ver = tuple([int(m.group(x)) for x in range(1, 5) if m.group(x)])
    except Exception as e:
        # we fall back to the old logic
        print(f"Failed to parse torch ver '{torch.version.__version__}' ({e})")
        if hasattr(torch, "compiler") and hasattr(torch.compiler, "disable"):
            return torch.compiler.disable
        else:
            return dummy_decorator
    else:
        if ver[0] < 2 or ver[:2] in [(2, 0), (2, 1)]: # 2.0.x, 2.1.x
            return dummy_decorator
        elif len(ver) == 4 and ver[3] > 20250401: # new nightly w/ fix
            return dummy_decorator
        elif len(ver) == 4: # old nightly
            return torch.compiler.disable
        elif ver[0] >= 2 and ver[1] >= 8: # 2.8.x or newer
            return dummy_decorator
        else: # old stable
            return torch.compiler.disable

# bypass or enable full torch compile depending on version
torch_compiler_disable = get_tc_decorator()

```

StrongerXi: Okay my stack is in the 2025/04/04 nightly and it adds [this to the config](https://github.com/pytorch/pytorch/commit/56d317d3a780a834c89b2c50ec33553379026947#diff-5f3e7e4866f93af05691e2382d2258cad959091492f900583a49e55266f5e897R163) (so that people can still opt-into the old way in case this new feature breaks). But that means we can conveniently `hasattr` on it rather than checking nightly date (which would help prevent breakage in the rare case my stack gets reverted).

I also found out about python's [builtin version handling](https://packaging.pypa.io/en/latest/version.html#usage).

So I updated my patch to use both of the above, which should make this decorator backward _and_ forward compatible, while delivering optimal performance. Lmk what you think:).

StrongerXi:
> We need bleeding edge pytorch for this?

Yes, ideally you just need to [`pip install` the nightly](https://pytorch.org/get-started/locally/).

This PR (along with latest PyTorch), basically allows `torch.compile` to capture and compile the entire model computation graph (rather than compiling a few isolated subgraphs), thereby delivering better perf out of the box. But the optimization _might_ not stop here, I'll try looking more into this to see if there's anything we can still squeeze out given this bigger graph.

There's also a bigger quest of better integration of `torch.compile` into ComfyUI we'll investigate into. If you guys have suggestions/pain-points to share, they'll be much appreciated:)

ops.py
```diff
@@ -6,15 +6,39 @@
...
+        return dummy_decorator
+    elif hasattr(torch._dynamo.config, "nontraceable_tensor_subclasses"):
+        # Here we are in the future but before the formal release
+        # (pre-release nightly). This allows early access to features
+        # mentioned in the branch above.
+        return dummy_decorator
+    else:
+        # Else we are in the middle, i.e., `torch.compiler.disable` exists
+        # and `torch.compile` hard errors without using it.
+        return torch.compiler.disable
```
city96, line 38: Looks good, think we can inline the `version.parse()` bits for the two static versions & shorten the comments to be one line to make it a bit more compact. Will grab nightly sometime tomorrow and give it a proper test.

Actually, if the `nontraceable_tensor_subclasses` is expected to be part of the release as well, then I think we could even skip checking the version and just check for that? What do you think

```py
if not hasattr(torch, "compiler") or not hasattr(torch.compiler, "disable"):
    return dummy_decorator # torch too old
elif hasattr(torch._dynamo.config, "nontraceable_tensor_subclasses"):
    return dummy_decorator # torch compile works
else:
    return torch.compiler.disable
```

StrongerXi: > Looks good, think we can inline the version.parse() bits for the two static versions & shorten the comments to be one line to make it a bit more compact. Will grab nightly sometime tomorrow and give it a proper test.

Updated.

> Actually, if the nontraceable_tensor_subclasses is expected to be part of the release as well, then I think we could even skip checking the version and just check for that? What do you think

The only concern is `nontraceable_tensor_subclasses` might get removed in the future (we don't generally don't maintain backward-compatibility for configs), so to make this feature both accessible-before-next-release, and future-compatible against potential removal of `nontraceable_tensor_subclasses`, I added the `hasattr(torch._dynamo.config, "nontraceable_tensor_subclasses")` check.

city96 (approved):
Alright, did a full test run on windows with the models and envs I had around, don't see any regressions or issues with the older versions.

Speedup is `2.09s/it` without and `1.38s/it` with torch compile (RTX 3080, Flux Q4_0), which is pretty decent overall.

Think we're good to merge. We can always follow up on other issues if any come up. Thanks again for helping with this!

Issue #285, opened 2025-06-06
[Announcement] GGUF conversion now available on colab.


Symbiomatrix:
@city96 
I don't really know where to put this since it's not quite a full solution, but it does seem to work without errors.
[The conversion process is difficult for a layman to follow.](https://github.com/city96/ComfyUI-GGUF/tree/main/tools) [Moreso because the instructions lead to errors.](https://github.com/city96/ComfyUI-GGUF/issues/251#issue-2993389786)
With a bit of llm magic and tinkering, I was able to get it to run on colab, grabbing the model file from a public huggingface repo. I do not know if it works as intended, and haven't tested the resulting gguf, but it certainly seemed to be doing the things it was supposed to. It runs on cpu, no gpu (T4) utilised, I'm not sure if that's possible, but anyway the process runs in colab for ~30 minutes so SEP.

https://colab.research.google.com/drive/145MdbMFFWdgJxl_CRN04mddEo3Qcxj2x?usp=sharing

Symbiomatrix:
This is a version which also deletes the source file mid write to conserve space, can even convert FP32 models (30gb+).
It also works with wan.

https://colab.research.google.com/drive/1b8H-Rd5cQo81AO7J5Nm34eryVCZHSOkN?usp=sharing

PR #286, opened 2025-06-06
[Efficiency improvement] GGUF conversion - file caching.

Symbiomatrix:
One line fix - ggufwriter will cache to file instead of living in ram.
It's a bit slower naturally, but I think it's unreasonable to demand - what is it, 24gb ram? I get crashes on 12gb at 54% - to store a full flux model in memory, when it's a one time operation.
As mentioned in #285 , this will allow running conversions even on low resource containers (colab, space presumably).

### DIFF
--- a/tools/convert.py
+++ b/tools/convert.py
@@ -307,7 +307,7 @@ def convert_file(path, dst_path=None, interact=True, overwrite=False):
             raise OSError("Output exists and overwriting is disabled!")
 
     # handle actual file
-    writer = gguf.GGUFWriter(path=None, arch=model_arch.arch)
+    writer = gguf.GGUFWriter(path=None, arch=model_arch.arch, use_temp_file=True) # Cache to file.
     writer.add_quantization_version(gguf.GGML_QUANT_VERSION)
     if ftype_gguf is not None:
         writer.add_file_type(ftype_gguf)


Symbiomatrix:
It's acting a bit weird, added some prints to see if it gets stuck during writing somewhere.

Symbiomatrix:
Double checked - seems to be working correctly, just that when you use a temp file the `write_tensors_to_file` function takes significantly longer - another 5 minutes, I reckon. Breaking it mid run will yield a corrupted file.

Symbiomatrix:
I added a model deletion flag, but colab doesn't appear to free any disk space whilst the process is running. The only way I could get it to work with FP32 is to return the writer after writing header, remove source model and then proceed with writing (per branch efficiency1). Semi manual hack but effective. Hf spaces seem to be even more limited at 50gb, colab at 70gb.

city96:
Sorry, I'm pretty swamped with IRL stuff lately so I can't really review or test stuff at the moment, but just some ideas:

There is a branch with an auto convert gradio here that might have some useful stuff you can reuse for the google colab space, including some of the logic: https://github.com/city96/ComfyUI-GGUF/blob/auto_convert/tools/tool_auto.py (+ the PR https://github.com/city96/ComfyUI-GGUF/pull/274 )

Some other nodes for this PR specifically:
 - I think adding `use_temp_file=True` would make sense to have behind a launch arg, with the description mentioning that it saves RAM but makes stuff slightly slower + adds some wear to the SSD I guess (also, possible to crash if the user has temp on the system drive and is running low on space? maybe keep the default and add a note to the readme). Dunno if this needs to have a version check to make sure old versions of `gguf-py` don't break if that flag is present.
 - There is this PR that I still need to look at. I want to add it to the gradio on the `auto_convert` branch then make it a HF space similar to gguf-my-repo but yeah, I had zero time to work on any of this in the past months sadly...
 https://github.com/ggml-org/llama.cpp/pull/12820 
- Deleting the input file is a definite no-go to have in the main repo. I don't like to have any kind of destructive operations that could result in people having to re-download files or even worse, losing custom finetuned models due to it.

PR #304, opened 2025-08-04
support lumina-image-2.0

spawner1145:
# models(neta lumina ggufs):

[modelscope](https://www.modelscope.cn/datasets/spawner/neta-lumina-gguf)

[huggingface](https://huggingface.co/neta-art/neta-lumina-gguf/tree/main)

complie with https://github.com/spawner1145/llama.cpp-backup.git


tools/convert.py
```diff
@@ -139,8 +139,19 @@ class ModelSD1(ModelTemplate):
...
+class ModelLumina2(ModelTemplate):
+    arch = "lumina2"
+    keys_detect = [
+        ("cap_embedder.1.weight", "context_refiner.0.attention.qkv.weight")
+    ]
+    keys_hiprec = [
+        # nn.parameter, can't load from BF16 ver
+        ".adaLN_modulation.1.bias",
+        ".adaLN_modulation.1.weight"
+    ]
```
city96, line 151: Are these meant to be kept in full precision? They seem to just be regular linear layers based on the shape.
The one you have on huggingface has them quantized, as the c++ code doesn't keep them in high precision either I don't think?

spawner1145:
hi @city96 ,thanks for your advice,I have changed the code by removing keys_hiprec and adding lines to the patch.

city96:
Cool, thanks. May want to re-convert the ones on HF with the changes, since as mentioned, the M and S ones are the same size.

image

Will merge now, can always send any fixes in a follow-up PR.

PR #330, opened 2025-09-08
Fix attempt to create a tensor with dtype "target"

blepping:
Currently setting `dequant_dtype` to "target" in the advanced GGUF unet loader will lead to a crash because we attempt to create an empty tensor with `dtype = "target"` which isn't a valid dtype.

This small change just uses the fallback dtype in that case. I don't know if this is the correct fix or not, but it seems to work.

### DIFF
--- a/ops.py
+++ b/ops.py
@@ -153,7 +153,7 @@ class GGMLLayer(torch.nn.Module):
         # Take into account space required for dequantizing the largest tensor
         if self.largest_layer:
             shape = getattr(self.weight, "tensor_shape", self.weight.shape)
-            dtype = self.dequant_dtype or torch.float16
+            dtype = (self.dequant_dtype != "target" and self.dequant_dtype) or torch.float16
             temp = torch.empty(*shape, device=torch.device("meta"), dtype=dtype)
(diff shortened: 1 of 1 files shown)


city96:
Ah, thanks. Not sure how this went unnoticed for this long.

I think theoretically you'd want the order of the 2 sides of that `and` swapped, since the default is for it to be `None`, so the expression doesn't have to evaluate it twice if you first check whether it's set or not in the default scenario:

```py
dtype = (self.dequant_dtype and self.dequant_dtype != "target") or torch.float16
```

Probably saves like, 2 cpu cycles overall at best lol but still.

blepping:
> Probably saves like, 2 cpu cycles overall at best lol but still.

Exactly the sort of likely pointless optimization I'd do, so how can I complain? :) Since Python strings are immutable, I'm pretty sure it would just be a pointer comparison and not actually scanning thorough the content... but changed!

city96:
Apparently it's ~20ns/call faster for the `None` case, I'm sure it's totally noticeable in real world scenarios while staring down the tqdm bar crawling along at 22s/it lol

image

PR #334, opened 2025-09-13
Fix dequant_dtype handling

blepping:
It's back!

Unfortunately, this actually has to be the way I originally had it and we'll just have to live with the 2ns performance difference. Otherwise `self.dequant_dtype and self.dequant_dtype != "target"` ends with a truthy value when those conditionals both succeed and short-circuits there, meaning the expression evaluates to `True` which doesn't work as a dtype passed to `torch.empty`.

### DIFF
--- a/ops.py
+++ b/ops.py
@@ -153,7 +153,7 @@ class GGMLLayer(torch.nn.Module):
         # Take into account space required for dequantizing the largest tensor
         if self.largest_layer:
             shape = getattr(self.weight, "tensor_shape", self.weight.shape)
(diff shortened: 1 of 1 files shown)


city96:
Tbh we should probably just make it a regular if so it's at least readable instead of optimizing performance in this case xD

```py
    if self.dequant_dtype is None or self.dequant_dtype == "target":
        dtype = torch.float16 # default or fallback
    else:
        dtype = self.dequant_dtype
```

blepping:
How about a compromise?

Issue #321, opened 2025-08-23
Please add gemma2 text encoder arch


btaskel:
Lumina2 requires gemma2 2b(5.2GB) as a text encoder, but it still takes up a lot of memory on some mobile devices. We hope to increase support for the gemma2 architecture🙏.

spawner1145:
Hi, btaskel. I have tested the GGUF format for the Gemma2-2B model using the code repository at https://github.com/spawner1145/ComfyUI-GGUF/tree/gemma2test and identified a critical issue: **significant quality loss** in generated outputs when using the GGUF-quantized version of the model.

pre-quantized models in https://huggingface.co/neta-art/gemma2-2b-gguf

## 1. Root Cause of Compatibility Issue
ComfyUI requires a specific metadata key named `"spiece_model"` to enable model integration with downstream tasks. However, this key is **not included** in the official implementation of llama.cpp. As a result, quantized GGUF models generated via standard llama.cpp workflows fail to be recognized and passed to subsequent processing steps in ComfyUI.


## 2. Solution Attempt: Custom Conversion Script
To resolve the metadata compatibility issue, I developed a custom conversion script that processes the official Safetensors files provided by ComfyUI. This script ensures the required `spiece_model` metadata is retained in the final GGUF file.

### How to Use the Conversion Script

Run the following command in the repository to convert a Gemma2-2B Safetensors file to GGUF format. The script supports several parameters to control the conversion process:

```bash
python tools/convert_gemma2.py --src gemma_2_2b_fp16.safetensors --dst gemma_2_2b_fp16.gguf --quantize q4_k
```

#### Parameter Explanations:

1.**`--src` (required)**- Specifies the path to the source Safetensors file containing the Gemma2-2B model
   - Example: `--src gemma_2_2b_fp16.safetensors`
   - This parameter is mandatory and must point to a valid existing file

2.**`--dst` (optional)**- Defines the output path and filename for the converted GGUF file
   - If not specified, the output filename will be generated automatically based on the source filename and quantization type (e.g., `gemma_2_2b_fp16-Q4_K.gguf`)
   - Example: `--dst custom_output_name.gguf`

3.**`--quantize` / `--quant` / `-q` (optional)**- Specifies the quantization type to apply during conversion
   - Available options:
     - Full precision: `f32`, `f16`, `bf16`
     - 8-bit quantization: `q8_0`
     - 4-bit quantization: `q4_0`, `q4_1`, `q4_k`
[...]

PR #346, opened 2025-10-02
Gemma2 text model support

spawner1145:
#321 

models in https://huggingface.co/neta-art/gemma2-2b-gguf

### DIFF
--- a/loader.py
+++ b/loader.py
@@ -181,7 +181,7 @@ GEMMA2_SD_MAP = {
     ".attn_norm.weight": ".input_layernorm.weight",
     ".post_attention_norm.weight": ".post_attention_layernorm.weight",
     ".post_ffw_norm.weight": ".post_feedforward_layernorm.weight",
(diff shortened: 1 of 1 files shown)


city96:
So I don't think we can merge this as-is, in part because it has a lot of random changes that look like AI slop to me (i.e. the logging messages with the checkmark emoji lol), but mainly because I don't think making custom model formats for text models supported by llama.cpp is acceptable.

We should only ever convert image models, and for the text models the conversion should be handled by mainline llama.cpp.

The tokenizer can be reconstructed from the metadata, though you need to do a tensor operation to actually make it usable, otherwise the norms will be messed up and you'll get bad results. I tried my best to try and implement it, though I couldn't get the tokenizer to be 1:1 (everything else is tested working with proper quants from huggingface) https://github.com/city96/ComfyUI-GGUF/pull/358

If you have the time, you could try and see if you can find the issue with the above PR, I'm pretty sure it's either due to `spm.trainer_spec` or due to the individual `piece.score` values. I left the test line in to load the quantized model but replace `sd["spiece_model"]` to test as the baseline. I can go into more detail, if you have any questions.

(Also, this PR would break UMT5 support, which already uses the tokenizer reconstruction logic)

Issue #379, opened 2025-12-08
GGUF size mismatch error on Mac


BuhleanJoe:
Hello,
when trying to run the z-image turbo workflow with guff-versions, I get this error in ComfyUi:

```
UnetLoaderGGUF
Error(s) in loading state_dict for NextDiT:
size mismatch for x_pad_token: copying a param with shape torch.Size([3840]) from checkpoint, the shape in current model is torch.Size([1, 3840]).
size mismatch for cap_pad_token: copying a param with shape torch.Size([3840]) from checkpoint, the shape in current model is torch.Size([1, 3840]).
```

ComfyUI version: 0.3.76

MacBook Air M2
Sonoma 14.1.1
RAM: 24 GB

Workflow:
<img width="1212" height="645" alt="Image" src="https://github.com/user-attachments/assets/f7f57431-b2ca-4870-822e-6d0f1df183b6" />

Thanks!

RubenGarcia:
I confirm the referenced patch fixes this issue.

PR #392, opened 2025-12-24
Fix lumina2 pad token shape mismatch for some GGUF conversions

vaclavmuller:
This PR fixes a shape mismatch when loading some lumina2 / NextDiT GGUF models
(e.g. Z-Image Turbo GGUF builds).

Some GGUF conversions store `x_pad_token` and `cap_pad_token` as 1D vectors
([D]) instead of the expected 2D shape ([1, D]), which causes
`load_state_dict` to fail.

The loader now:
- ensures a robust fallback shape when `orig_shape` metadata is missing
- reshapes lumina2 pad tokens to (1, D) when needed

Tested with:
https://huggingface.co/leejet/Z-Image-Turbo-GGUF

Addresses #379

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/loader.py
+++ b/loader.py
@@ -106,13 +106,24 @@ def gguf_sd_loader(path, handle_prefix="model.diffusion_model.", return_arch=Fal
             torch_tensor = torch.from_numpy(tensor.data) # mmap
 
         shape = get_orig_shape(reader, tensor_name)
+
+        # fallback shape when original shape metadata is missing in GGUF
         if shape is None:
-            shape = torch.Size(tuple(int(v) for v in reversed(tensor.shape)))
+            # tensor.shape may be None in some cases -> use torch_tensor shape as fallback
+            raw_shape = tensor.shape if tensor.shape is not None else torch_tensor.shape
+            shape = torch.Size(tuple(int(v) for v in reversed(raw_shape)))
+
             # Workaround for stable-diffusion.cpp SDXL detection.
             if compat == "sd.cpp" and arch_str == "sdxl":
                 if any([tensor_name.endswith(x) for x in (".proj_in.weight", ".proj_out.weight")]):
                     while len(shape) > 2 and shape[-1] == 1:
                         shape = shape[:-1]
+
+        # --- workaround for lumina2 / NextDiT pad tokens (Z-Image Turbo) ---
+        if arch_str == "lumina2" and sd_key in ("x_pad_token", "cap_pad_token"):
+            if len(shape) == 1:
+                shape = torch.Size((1, shape[0]))
+        # ------------------------------------------------------------------
 
         # add to state dict
         if tensor.tensor_type in {gguf.GGMLQuantizationType.F32, gguf.GGMLQuantizationType.F16}:


altarhex:
thank you! this fixed the issue I was having

RubenGarcia:
I'd also like this merged.

RubenGarcia (approved):

ItsG33k (approved):

hinablue:
Thank you! That solved my problem.

RubenGarcia:
@city96 Is there anything else you need done here?

qskousen:
I don't pretend to understand all of what is going on here, but I've found that this only causes an error if these two layers are in BF16; if they are F32 or F16 they seem to work fine. Not sure why that is. Or, I may be experiencing an unrelated issue.

patientx:
this should be on the main.

RubenGarcia:
While this fixes loading for zimage turbo, I am getting similar errors for zimage base.
@vaclavmuller, can you have a look?
These are the affected models:
https://huggingface.co/babakarto/z-image-base-gguf/resolve/main/z_image_base_BF16.gguf
https://huggingface.co/babakarto/z-image-base-gguf/resolve/main/z_image_base_Q8_0.gguf

z_image_base_Q8_0.gguf:
The expanded size of the tensor (3840) must match the existing size (4080) at non-singleton dimension 1.  Target sizes: [3840, 3840].  Tensor sizes: [3840, 4080]

z_image_base_BF16.gguf
The expanded size of the tensor (3840) must match the existing size (7680) at non-singleton dimension 1.  Target sizes: [3840, 3840].  Tensor sizes: [3840, 7680]

82sound-jpg:
> While this fixes loading for zimage turbo, I am getting similar errors for zimage base. @vaclavmuller, can you have a look? These are the affected models: https://huggingface.co/babakarto/z-image-base-gguf/resolve/main/z_image_base_BF16.gguf https://huggingface.co/babakarto/z-image-base-gguf/resolve/main/z_image_base_Q8_0.gguf
> 
> z_image_base_Q8_0.gguf: The expanded size of the tensor (3840) must match the existing size (4080) at non-singleton dimension 1. Target sizes: [3840, 3840]. Tensor sizes: [3840, 4080]
> 
> z_image_base_BF16.gguf The expanded size of the tensor (3840) must match the existing size (7680) at non-singleton dimension 1. Target sizes: [3840, 3840]. Tensor sizes: [3840, 7680]

我也是这个问题

vaclavmuller:
Thanks for testing and for reporting this.

The fix in this PR only addresses the pad token shape mismatch that occurs with some Z-Image Turbo GGUF conversions (`x_pad_token` / `cap_pad_token` stored as [D] instead of [1, D]). From the error messages you posted, the issue with the Z-Image Base models looks different.

These errors:

Q8_0:
The expanded size of the tensor (3840) must match the existing size (4080) at non-singleton dimension 1

BF16:
The expanded size of the tensor (3840) must match the existing size (7680) at non-singleton dimension 1

suggest that the tensor shapes stored in the GGUF file do not match what the current NextDiT / lumina2 implementation in ComfyUI expects. That is most likely related either to:
- differences in the GGUF conversion script used for those models, or
- missing / incorrect `orig_shape` metadata in the GGUF file.

In other words, this is probably not the same issue that this PR fixes.

For transparency: I am not a core developer of this repository. I ran into the Turbo issue while using ComfyUI and worked through the debugging process with the help of ChatGPT, which helped me understand the loader code and produce a minimal fix for that specific problem.

If you want to investigate the Z-Image Base issue further, I would recommend:
1. checking the tensor names and shapes inside the GGUF file,
2. comparing them with the shapes expected by the lumina2 / NextDiT model in ComfyUI,
3. verifying whether the conversion process added correct `comfy.gguf.orig_shape.*` metadata.

If you are not familiar with the loader code, tools like ChatGPT can actually be very helpful for exploring the code and reasoning about shape mismatches like this. That’s essentially how I approached the Turbo issue as well.

If someone can identify which specific tensor is triggering the mismatch (the stack trace usually shows it), that would be a good starting point to determine whether this needs:
- a loader-side workaround similar to the Turbo fix, or
- a correction in the GGUF conversion process.

PR #399, opened 2026-01-07
Added support to load model config from Metadata.

vantagewithai:
The Diffusion Loader in ComfyUI can read the model config directly from the safetensors metadata. I’ve added the same support, so newer models like LTX2 are handled correctly, since ComfyUI loads the model configuration for LTX2 from the safetensors file header metadata.

PR #402, opened 2026-01-09
Add Gemma3 12B Support

jarz76:
Trying to add support for Gemma 3 12B GGUF, it can be used with the `DualClipLoader (GGUF)` node.

CLIP 1 is a Gemma 3 GGUF, and CLIP 2 uses embedding connectors from: https://huggingface.co/Kijai/LTXV2_comfy/tree/main/text_encoders

Note: The connectors from @kijai seem to come from the distilled models, and when testing it shows different results compared to connectors extracted from the dev models. ~~I’ve uploaded the connectors-dev here: https://huggingface.co/jayn7/LTXV2/tree/main if anyone want try it~~. Kijai has updated the repo and now provides both dev and distilled connectors.

chrome_iVsVdBcPKv

~~This approach uses the Gemma 3 `tokenizer.model` (4.5MB) file directly, instead of attempting to recreate tokenizer from metadata. It loads and searches for `tokenizer.model` or `gemma3-tokenizer.model` inside `ComfyUI/models/text_encoders` folder.
The tokenizer can be found here: https://huggingface.co/google/gemma-3-12b-it/tree/main~~

Edit: We no longer need `tokenizer.model`, the approach is the same as the others, it will attempt to recreate the tokenizer from metadata. https://github.com/city96/ComfyUI-GGUF/pull/402#issuecomment-3735608909

GGUF quants tested so far, but as long as they contain the required metadata, any release should work fine:
https://huggingface.co/unsloth/gemma-3-12b-it-GGUF - IQ4_XS
https://huggingface.co/bartowski/google_gemma-3-12b-it-GGUF - Q8 & BF16
https://huggingface.co/google/gemma-3-12b-it-qat-q4_0-gguf

Some example results (workflow embedded)

LTX 2.0 DEV T2V
| # | Description |  Result (video) |
|---|--------------|----------------|
| 1 | ComfyUI LTXV Audio Text Encoder Node (gemma_3_12B_it_bf16.safetensors) | <video src="https://github.com/user-attachments/assets/ba081abe-598b-4510-9474-fd938b38ae61" controls muted loop></video> |
| 2 | DualClipLoader GGUF Node (Gemma3 GGUF Q8 + embedding connectors extracted from dev models) | <video src="https://github.com/user-attachments/assets/795ef20f-67e1-4cdf-a5cc-67c0baeea555" controls muted loop></video> |
[...]

PR #404, opened 2026-01-10
Adding Gemma3 Support (correctly loading tokenizer and mmproj, alongside existing GGUFs).

scottmudge:
This is a concise implementation for loading the Gemma3 tokenizer, with the correct configuration values and an appropriate method of loading the tokenizer (`spiece` model) and mmproj alongside existing GGUF quants available on huggingface.

I've extracted the tokenizer into a standalone model, along with the mmproj. **Note** that these two supplementary models are *required*:

https://huggingface.co/smhf72/gemma-3-12b-it-extras-comfy

These two models should be placed in the `models/clip` directory, and they are automatically loaded when a `gemma3` architecture is detected.

It works as expected with these GGUFs:

* https://huggingface.co/unsloth/gemma-3-12b-it-GGUF
* https://huggingface.co/mradermacher/gemma-3-12b-it-norm-preserved-biprojected-abliterated-GGUF

and others.

---

EDIT:

If testing, this expects https://github.com/city96/ComfyUI-GGUF/pull/399 to be merged first.

### DIFF
--- a/loader.py
+++ b/loader.py
@@ -5,12 +5,13 @@ import torch
 import gguf
 import re
 import os
+from pathlib import Path
 
 from .ops import GGMLTensor
 from .dequant import is_quantized, dequantize_tensor
 
 IMG_ARCH_LIST = {"flux", "sd1", "sdxl", "sd3", "aura", "hidream", "cosmos", "ltxv", "hyvid", "wan", "lumina2", "qwen_image"}
-TXT_ARCH_LIST = {"t5", "t5encoder", "llama", "qwen2vl", "qwen3", "qwen3vl"}
+TXT_ARCH_LIST = {"t5", "t5encoder", "llama", "qwen2vl", "qwen3", "qwen3vl", "gemma3"}
 VIS_TYPE_LIST = {"clip-vision", "mmproj"}
 
 def get_orig_shape(reader, tensor_name):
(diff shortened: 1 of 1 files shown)


zwukong:
cool , but is mmproj.gguf necessary? that pr doesn't need that

scottmudge:
Yes it is necessary for visual reasoning. The original model distributed with the ComfyUI Gemma 3 safetensor release included the mmproj tensors.

zwukong:
OK，i tried i2v without mmproj,seems work

scottmudge:
Well by all means if you don't want it for whatever reason, don't use it. But I figured they included it for a reason. This PR should work fine without it (as in it won't raise an exception), but it will issue a warning. 

In the ComfyUI-LTX repo's example workflows, they use a prompt "enhancer" node, which uses the image as an input to Gemma 3 and asks it to improve the prompt based on what is in the input image.

You cannot do this without the mmproj model, for example. The model won't know what the image tokens are.

zwukong:
i know what you mean, something like VL image edit model all need  that. But native comfyui clip node can not give string as result, so VL model can not be used very efficiently

scottmudge:
That is true, however other custom nodes may want to make use of visual reasoning using the GGUF loader, so allowing the mmproj file to be loaded, as it already does with Qwen, made sense. 

When it comes to making things easier for the user, it's generally better to include things by default, so they won't run into issues later if they suddenly need or want it (perhaps without even knowing what it is).

Andy92j:
I always get this error: not enough values to unpack (expected 3, got 2)
I've tried various gguf's (mostly the Q2 version, though)

scottmudge:
Sorry, you need to merge https://github.com/city96/ComfyUI-GGUF/pull/399 first, then merge this PR.

https://github.com/city96/ComfyUI-GGUF/pull/399 is needed for loading LTX-2's models (not the CLIP) as GGUF anyway.

frost-byte:
> Sorry, you need to merge #399 first, then merge this PR.
> 
> #399 is needed for loading LTX-2's models (not the CLIP) as GGUF anyway.

Is #402 also needed? I was having an issue when using a gemma gguf with those prs merged - it was just garbled output. I have already downloaded all of the files mentioned, including the .mmproj. I assume they just need to reside in the text_encoders folder. (I did try the fp8_e4m3n version and that did work.
Actually, I guess the workflow that was able to run was t2v, trying i2v with fp8_e4m3n produced the same garbled output. So I must have something else incorrect.

zwukong:
i don't think your mothed is right. ram cost is huge, gemma3 q4km 7G cost more than 50G ram to load. the other pr do not have that problem

PR #433, opened 2026-04-06
feat: add torch dequantization for IQ1_S, IQ1_M, IQ2_XXS, IQ2_S, IQ3_XXS, IQ3_S

### LINKED ISSUE #405: More IQ quant types and fix for torch.compile
Hi, I've implemented more quant types `IQ1_S, IQ1_M, IQ2_XXS, IQ2_S, IQ3_XXS, IQ3_S` in torch at https://github.com/woct0rdho/transformers-qwen3-moe-fused/blob/master/qwen3_moe_fused/quantize_gguf/dequant.py

Also I've found that `torch.compile` cannot handle some complicated `view` operations, and it silently gives wrong numbers or nan. I've added `view_float16` in dequant functions like Q6_K to work around it, and rewritten the IQ4_XS dequant function. Now all dequant functions work with `torch.compile` correctly. Later we should find some minimal reproducers and report to PyTorch.

There is a unit test at https://github.com/woct0rdho/transformers-qwen3-moe-fused/blob/master/test_gguf_dequant.py

I'm not sure if we're interested in quantizing diffusion models into IQ3 or smaller quants, but this at least allows us to load some existing LLMs as text encoders, especially for Unsloth UD quants. I can make a PR to this repo if needed.

octo-patch:
Closes #405

## Problem

Lower IQ quant types (IQ1_S, IQ1_M, IQ2_XXS, IQ2_S, IQ3_XXS, IQ3_S) currently fall back to the slow numpy path in `dequantize_tensor`:

```
# this is incredibly slow
tqdm.write(f"Falling back to numpy dequant for qtype: ...")
new = gguf.quants.dequantize(tensor.cpu().numpy(), qtype)
```

This makes loading models quantized with these formats (e.g. Unsloth UD quants used as text encoders) very slow.

## Solution

Implements native PyTorch dequantization functions for all six missing IQ quant types, using the quantization grids already provided by the `gguf` Python package.

The implementations follow the same pattern as the existing `IQ4_NL` and `IQ4_XS` dequantizers and were cross-referenced against the numpy reference implementation in `gguf.quants`.

## Testing

All new functions are verified against `gguf.quants.dequantize()` (the numpy reference):

```
Q8_0:    max_diff=0.000000 ✓
Q6_K:    max_diff=0.000000 ✓
IQ4_NL:  max_diff=0.000000 ✓
IQ4_XS:  max_diff=0.000000 ✓
IQ3_S:   max_diff=0.000000 ✓
IQ3_XXS: max_diff=0.000000 ✓
IQ2_S:   max_diff=0.000000 ✓
IQ2_XXS: max_diff=0.000000 ✓
IQ1_M:   max_diff=0.000000 ✓
IQ1_S:   max_diff=0.000000 ✓
```

### DIFF
--- a/dequant.py
+++ b/dequant.py
@@ -1,5 +1,6 @@
 # (c) City96 || Apache-2.0 (apache.org/licenses/LICENSE-2.0)
 import gguf
+import numpy as np
 import torch
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


SSJGabraham:
Tested this and it works well for IQ3_S in my setup.

octo-patch:
Thanks for testing @SSJGabraham! Great to hear IQ3_S is working well. If you get a chance to try any of the other quant types (IQ1_S, IQ1_M, IQ2_XXS, IQ2_S, IQ3_XXS), would love to hear how those work for you too.

SSJGabraham:
@octo-patch One thing I noticed is that when I first ran a test, I thought I was using a IQ3_S, but it was actually IQ3_M, which if I am not mistaken was not listed in your new methods or in your summary above, but it still worked without reverting to the slow numpy path. I didn't notice a slowdown and I did not get a stream of messages in my console about it. When I switched to the proper IQ3_S model I intended to use, performance was identical.

Was this expected?

PR #445, opened 2026-05-10
nodes: skip self.model.to(device_to) walk in unpatch_model on mmap'd weights

### LINKED ISSUE #444: [Bug Report] Windows fatal exception: access violation when node has multiple outputs
When using the latest comfyui-gguf 2.0.0 on Windows, I get a crash:
plaintext
Windows fatal exception: access violation
Reproduce step
Modify the UnetLoaderGGUF node to return two outputs: (MODEL, STRING)
After generating images, ComfyUI will automatically trigger unpatch_model
Crash occurs inside GGUFModelPatcher.unpatch_model
Root cause
GGUF quantized weights use mmap memory mapping.
When passing unpatch_weights=True to super().unpatch_model,
the base model patcher tries to operate directly on GGUF raw tensor memory,
causing Windows memory access violation and instant crash.
Fix solution
Only one line needs to be modified:
Original line:
python:
return super().unpatch_model(device_to=device_to, unpatch_weights=unpatch_weights)
Fixed line:
python:
return super().unpatch_model(device_to=device_to, unpatch_weights=False)
Why this fix works
Set unpatch_weights=False to avoid the base patcher touching GGUF mmap weight memory.
Model can still unload normally when switching models, no memory leak.
No crash, all original GGUF functions keep working perfectly.
This is a regression bug in 2.0.0, the fix was already applied in old version 1.1.6 but missed in the refactored 2.0.0 code.

Booyaka101:
Resolves #444.

`GGUFModelPatcher.unpatch_model` forwards to base `ModelPatcher.unpatch_model`, which finishes its `unpatch_weights=True` block with `self.model.to(device_to)`. That walk uses `nn.Module.to` and iterates every parameter, ignoring `comfy_cast_weights = True` on `GGMLLayer`. The walk hits the still-mmap'd quantized tensors and faults with `EXCEPTION_ACCESS_VIOLATION` on Windows. Repro: any workflow that triggers a cached-patcher flush during sampling (e.g. a `UnetLoaderGGUF` wired into a multi-output node).

The cleanest fix is upstream, since the same conflict will hit any custom op that uses `comfy_cast_weights` with tensors that cannot survive `nn.Module.to`. Tracked at Comfy-Org/ComfyUI#14142.

In the meantime this override inlines base's `unpatch_weights=True` block and skips just that one line. The block has to run in full because `partially_load` reads `model_loaded_weight_memory` immediately after `unpatch_model` and short-circuits `self.load()` if it is nonzero - so a partial mirror (the previous attempt at this PR) left `.patches` cleared without re-attaching anything on LoRA strength changes (caught by @romybaby on this thread). The lines this override now runs map one-for-one to comfy's `model_patcher.py` body:

- `eject_model`
- `unpatch_hooks`, `unpin_all_weights`
- lowvram cleanup via `move_weight_functions` / `wipe_lowvram_weight`
- non-quantized backup restoration
- `current_weight_patches_uuid = None`, `backup.clear()`
- `model_loaded_weight_memory = 0`, `model_offload_buffer_memory = 0`
- `comfy_patched_weights` deletion per module
- final `super().unpatch_model(unpatch_weights=False)` for `eject_model` (idempotent) and `object_patches_backup` restoration

Drift risk noted: if upstream adds something to that block, this override needs a sync. Easier to spot since the structure now matches line for line.

Credit @stonerabit for the original repro and diagnosis on #444, and @romybaby for catching the LoRA strength-change regression in the previous iteration.

### DIFF
--- a/nodes.py
+++ b/nodes.py
@@ -74,8 +74,29 @@ class GGUFModelPatcher(comfy.model_patcher.ModelPatcher):
                 patches = getattr(p, "patches", [])
                 if len(patches) > 0:
                     p.patches = []
-        # TODO: Find another way to not unload after patches
-        return super().unpatch_model(device_to=device_to, unpatch_weights=unpatch_weights)
+            # Restore non-quantized backups locally so we can short-circuit
+            # super()'s weight-unpatch block — its self.model.to(device_to)
+            # walk crashes with a Windows access violation on the still-mmap'd
+            # quantized tensors (#444). The maintainer's previous mitigation
+            # (Sep 2024 commit 6dbb4ba, reverted in 717a0e1) sidestepped the
+            # walk via device_to=None but skipped backup restoration, breaking
+            # LoRA correctness on subsequent runs (temp_weight in
+            # patch_weight_to_device reads the still-patched current weight).
+            # Doing the restore ourselves keeps non-quantized weights
+            # round-tripping correctly.
+            keys = list(self.backup.keys())
+            for k in keys:
+                bk = self.backup[k]
+                if bk.inplace_update:
+                    comfy.utils.copy_to_param(self.model, k, bk.weight)
+                else:
+                    comfy.utils.set_attr_param(self.model, k, bk.weight)
+            self.model.current_weight_patches_uuid = None
+            self.backup.clear()
+            for m in self.model.modules():
+                if hasattr(m, "comfy_patched_weights"):
+                    del m.comfy_patched_weights
+        return super().unpatch_model(device_to=device_to, unpatch_weights=False)
 
 
     def pin_weight_to_device(self, key):
--- a/nodes.py
+++ b/nodes.py
@@ -67,40 +67,31 @@ class GGUFModelPatcher(comfy.model_patcher.ModelPatcher):
             comfy.utils.set_attr_param(self.model, key, out_weight)
 
     def unpatch_model(self, device_to=None, unpatch_weights=True):
-        self.eject_model()
         if unpatch_weights:
             for p in self.model.parameters():
                 if is_torch_compatible(p):
                     continue
-                if len(getattr(p, "patches", [])) > 0:
+                patches = getattr(p, "patches", [])
+                if len(patches) > 0:
                     p.patches = []
-            # Mirror of base unpatch_model's unpatch_weights block, skipping the
-            # self.model.to(device_to) walk that faults on mmap'd quantized
-            # tensors (#444). Tracking upstream at Comfy-Org/ComfyUI#14142.
-            self.unpatch_hooks()
-            self.unpin_all_weights()
-            if self.model.model_lowvram:
+            if device_to is not None and self.model.model_lowvram:
                 for m in self.model.modules():
                     comfy.model_patcher.move_weight_functions(m, device_to)
-                    comfy.model_patcher.wipe_lowvram_weight(m)
-                self.model.model_lowvram = False
-                self.model.lowvram_patch_counter = 0
-            for k in list(self.backup.keys()):
-                bk = self.backup[k]
-                if bk.inplace_update:
-                    comfy.utils.copy_to_param(self.model, k, bk.weight)
-                else:
-                    comfy.utils.set_attr_param(self.model, k, bk.weight)
-            self.model.current_weight_patches_uuid = None
-            self.backup.clear()
-            if device_to is not None:
-                self.model.device = device_to
-            self.model.model_loaded_weight_memory = 0
-            self.model.model_offload_buffer_memory = 0
-            for m in self.model.modules():
-                if hasattr(m, "comfy_patched_weights"):
-                    del m.comfy_patched_weights
-        return super().unpatch_model(device_to=device_to, unpatch_weights=False)
+        # device_to=None makes base skip its nn.Module.to walk, which touches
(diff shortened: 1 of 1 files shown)


romybaby:
Confirmed working on real CUDA. RTX 3060 (12GB), Windows 10, ComfyUI 0.22.2, PyTorch 2.8.0+cu129. Running Flux 2 Klein 9b, with both TE and diffusion model as GGUFs. Was facing an issue where any attempt to unload GGUF model weights (either manually or automatically, in my case triggered by normal between-run model caching) led to the same "Windows fatal exception: access violation" fault described above. After applying this patch, both manual and automatic unloads work and repeated runs are stable. LoRAs produce identical output across multiple runs with the same seed.

romybaby:
After further usage, one edge case to report: changing LoRA strength between runs produces output as if no LoRA is applied at all. Appears the quantized tensors' .patches are cleared correctly on unpatch, but not re-attached on the subsequent in-place re-patch since the model stays resident (unsure about this, though). A full model reload between strength changes fixes it.

Booyaka101:
@romybaby thanks for catching this, the regression is real and caused by the previous version of this PR.

What happened: forwarding `unpatch_weights=False` to base `unpatch_model` dodged the `self.model.to(device_to)` crash, but it also skipped the `model_loaded_weight_memory = 0` and `model_offload_buffer_memory = 0` lines that live in the same block. `partially_load` reads `model_loaded_weight_memory` right after `unpatch_model` and short-circuits `self.load()` if it is nonzero, so on a LoRA strength change the quantized `.patches` got cleared but never re-attached. Result: no LoRA applied on the next run, exactly what you saw. Full model reload worked because cold load does not go through that check.

Just pushed 32414a1 which inlines the whole `unpatch_weights=True` body minus the offending device walk. That keeps the memory counters zeroed so `load()` runs and re-attaches `.patches` at the new strength. The upstream feature request that would let this override collapse to a one-liner is at Comfy-Org/ComfyUI#14142.

Worth retesting your strength-change workflow with this commit if you have a moment.

romybaby:
Thanks for the quick response! The commit works perfectly on my end, changing strengths between runs now correctly reapplies the LoRA at the new strength level.

woodenriver05:
Additional unload-correctness data point for the current PR shape (single host, single A/B run).

On ComfyUI 0.33.1 + PyTorch 2.13.0+cu130, I tested a downstream override with the same relevant behavior as `32414a1`: run the base `unpatch_weights=True` cleanup and zero the memory counters, but omit the final `self.model.to(device_to)` walk.

With a fully loaded H3 Q5 GGUF text encoder:

- encoder load: `19039.68 MB`
- after the mirrored unload block, the following Q5 DiT saw only `6861.54 MB` available
- DiT load: partial, with `16766.75 MB` offloaded to CPU
- the run was stopped at 29m49s without output

Restoring the upstream delegation (including the final device move) on the same graph produced:

- DiT available memory: `25918.22 MB`
- DiT: `loaded completely; 23276.97 MB loaded, full load: True`

The upstream run did not reproduce the access violation on this current stack. That is only a current-stack non-reproduction; it does not invalidate the older Windows/PyTorch crash reports.

The important result for this PR is that resetting `model_loaded_weight_memory` and `model_offload_buffer_memory` without actually moving a fully loaded GGUF model can leave physical CUDA residency inconsistent with ComfyUI's accounting. The current implementation fixes the LoRA reload regression, but a full detach/unload can still retain roughly the entire prior model and force the next model into CPU offload.

This suggests the upstream fix needs to distinguish the `partially_load` hot path from a real detach/full-unload path, or provide a GGUF-safe way to achieve the requested final device residency. Simply omitting the walk for both paths is not unload-correct.

Booyaka101:
Confirmed on my end, and on a full detach it's worse than the partial offload you saw: ComfyUI v0.33.1, torch 2.11.0+cu130, Windows 11, RTX 4090, Wan2.2-I2V-A14B Q5_K_M (10427MB). With 32414a1 a full load then detach left cuda_alloc at 10512MB with both counters zeroed, and free_memory after loading a second copy left 20475MB stranded. Stock frees both to 0.

b543b78 drops the mirrored block. The override now passes device_to=None to super so base runs its whole unpatch block and skips only the nn.Module.to walk, then explicitly moves the tensors that report a different device and reassigns them with set_attr_param. Host-resident tensors are never touched, so mmap-backed memory only ever gets read. That covers your partially_load vs detach split by device residency instead of call site: on a real unload the moved set is exactly what has to leave the GPU, on the hot path nothing extra happens. It's also the device_to=None idea from 6dbb4ba that 717a0e1 reverted for breaking VRAM estimation, with the explicit move covering the accounting half.

Re-ran the matrix on b543b78: detach frees 10512MB to 0, double-load free_memory 20475MB to 0, LoRA strength 0.25 to 0.75 to detach matches main's patch state exactly, and the fresh, lowvram and cache-flush unpatch paths all complete clean.

On the original crash, same result as you: no repro on this stack across five scenario shapes. Probing the mechanism directly, .to() over a still-mmap'd tensor is fine here, but a single in-place write into one is an instant access violation. torch.from_numpy on gguf's read-only mapping gives a tensor torch treats as writable, which is the numpy warning loader.py suppresses. So on current stacks the part that matters is that unpatch never writes mapped memory in place, and this shape preserves that.

Issue #464, opened 2026-06-24
Please add gguf‑quantized support for "Krea‑2"


makisekurisu-jp:
diffusion_models:
https://huggingface.co/realrebelai/KREA-2_GGUFs/tree/main/TURBO
https://huggingface.co/realrebelai/KREA-2_GGUFs/tree/main/BASE

text_encoders:
https://huggingface.co/unsloth/Qwen3-VL-4B-Instruct-GGUF

<img width="1920" height="1000" alt="Image" src="https://github.com/user-attachments/assets/9989c98c-c7dd-41b9-93fb-441176cb70e6" />

RealRebelAI:
My gguf nodes are a forked version of these for krea-2, if anyone knows how to merge my loaders py into the normal gguf repo then we could all already have it. I coded it in last night. Its on my repo but i dont know how to use github at all to merge it myself. This site is like a foreign language for some reason to me

m8rr:
https://github.com/city96/ComfyUI-GGUF/issues/463#issuecomment-4786493967
https://github.com/m8rr/ComfyUI-GGUF/tree/Dynamic-VRAM

@RealRebelAI Your GGUF works fine in my environment, though I'm not sure if the calculations are correct. The GGUF from the link provided by @makisekurisu-jp throws a [WARNING] unet unexpected: ['last.down.weight', 'last.up.weight'], but it does work.

And ComfyUI's official BF16 doesn't have ['last.down.weight', 'last.up.weight'], so it seems safe to ignore.

ideogram4, boogu, krea2 / UNet and TE both function properly using GGUF. 
While Gemma-4 E2B appears to have some problems, it operates well in my specific environment, including vision recognition.

makisekurisu-jp:
> https://github.com/city96/ComfyUI-GGUF/issues/463#issuecomment-4786493967
> https://github.com/m8rr/ComfyUI-GGUF/tree/Dynamic-VRAM
> 
> @RealRebelAI Your GGUF works fine in my environment, though I'm not sure if the calculations are correct. The GGUF from the link provided by @makisekurisu-jp throws a [WARNING] unet unexpected: ['last.down.weight', 'last.up.weight'], but it does work.
> 
> And ComfyUI's official BF16 doesn't have ['last.down.weight', 'last.up.weight'], so it seems safe to ignore.
> 
> ideogram4, boogu, krea2 / UNet and TE both function properly using GGUF. 
[...]

PR #473, opened 2026-08-05
MiniMax H3 mmproj support for Qwen3 VL 32B GGUFs

Nif00:


allanmeng:
I hit the same problem from the Krea2 side (#464) and traced it to the Qwen3-VL mmproj
mapping. Your branch is the only one that wires `qwen3vl` into `gguf_mmproj_loader` at
all — but as written the DeepStack keys can't land, so Krea2 still fails. Two defects:

1. `CLIP_VISION_QWEN3_MAP` maps `"v.deepstast."`. The real tensors in
   `Qwen3-VL-4B-Instruct-mmproj-bf16.gguf` are `v.deepstack.{5,11,17}.{norm,fc1,fc2}.{weight,bias}`
   (18 tensors, verified with GGUFReader). `"deepstast"` never matches, so the DeepStack
   tensors pass through unrename­d.

2. The layer-index remap (`ck.5.`→`st.0.`, `ck.16.`→`st.1.`, …) is bound to a `ck.` prefix
   that doesn't exist in the file. Even with (1) fixed, you'd emit
   `model.visual.deepstack_merger_list.5/11/17.*`, but `comfy.sd.detect_te_model()` only
   probes `model.visual.deepstack_merger_list.0.norm.weight` — so it still falls back to
   `QWEN3_4B` and Krea2 still raises "expects 12x2560=30720 features but got 2560".

I have a small fix (`loader.py` only, +53/-3) that derives the index by sorting the layer
ids present in the file instead of hardcoding 4B/8B, plus an offline check against
`qwen3vl_4b_fp8_scaled.safetensors` (315/315 keys, cos ≈ 1.000) and a B580 hardware run.
Happy to rebase it on top of your branch if you'd rather keep everything in this PR —
otherwise I'll open it separately as a focused fix.

